PR #1, opened 2022-05-10
fix reshape to match matlab

jdtuck:
in the tensor class we have issues when using reshape as it does not match MATLAB TensorToolbox, not sure how far this propagates, but here is one fix that I need to do to match Matlab

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *~
 .*~
 pyttb.egg-info/
+dist/
 .pytest_cache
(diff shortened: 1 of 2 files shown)


jdtuck:
This is a partial fix for #3

dmdunla (approved):
Approving changes. This fixes the constructor, `from_data`, but the use of reshape will need to be thoroughly checked in the `tensor` class before further use.

PR #31, opened 2022-07-06
Update nvecs to use tenmat.

dmdunla:


dmdunla:
Update nvecs to use tenmat. Closes #25.

Issue #2, opened 2022-05-10
collapse in tensor class errors, not well documented on what the format of dims is. 


jdtuck:
When using `collapse` method in the `tensor` class, the parameter `dims` is not clear and leads to errors.

PR #32, opened 2022-07-06
Full implementation of collapse. Required implementation of tensor.fr…

dmdunla:
…om_tensor_type for tenmat objects. Updated tensor tests.


dmdunla:
Closes #2

PR #49, opened 2023-01-27
Compatibility with numpy 1.24.x

brian-kelley:
Resolve some issues involving deprecations and changes in NumPy 1.24.

- Type aliases ``numpy.float`` and ``numpy.int`` were deprecated in 1.20 and removed in 1.24, so replace with built-ins ``float`` and ``int``. These are identical replacements: https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated
- Creating ragged numpy arrays (numpy calls them inhomogeneous) is no longer allowed. To get an ordered set of numpy arrays with different dimensions (e.g. the ``vector`` argument of the ``ttv`` methods) allow plain lists to be used. This is a backward-compatible change - as long as you can construct ``vector`` as a numpy.ndarray, you can still pass it.

For an example of the second issue, in numpy 1.24:
```
>>> import numpy as np
>>> A = np.zeros(5)
>>> B = np.zeros(7)
>>> A
array([0., 0., 0., 0., 0.])
>>> B
array([0., 0., 0., 0., 0., 0., 0.])
>>> mylist = [A, B]
>>> mylist
[array([0., 0., 0., 0., 0.]), array([0., 0., 0., 0., 0., 0., 0.])]
>>> myarr = np.array(mylist)
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous
shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.
```
Before, this was allowed and you would simply get a 1D np.ndarray where the elements are themselves np.ndarrays, but now it checks against this.

Tested changes on two different python envs: a python 3.6.8 with numpy 1.19.5, and python 3.9.15 with 1.24.1. So that's one before the relevant numpy deprecations, and one after.
- Pytest runs cleanly
- Ran Jon Berry's TTB demo. It exercises sparse, dense and Kruskal tensor construction, MTTKRP and CP-ALS

### DIFF
--- a/pyttb/ktensor.py
+++ b/pyttb/ktensor.py
@@ -1806,7 +1806,7 @@ class ktensor(object):
         elif isinstance(dims, (float, int)):
             dims = np.array([dims])
 
-        assert len(vector) > 0, 'sptensor.ttv: vector must be nonempty'
+        assert len(vector) > 0, 'ktensor.ttv: vector must be nonempty'
 
         # Check that vector is a list of vectors, if not place single vector as element in list
         if isinstance(vector[0], (int, float, np.int_, np.float_)):
--- a/pyttb/ktensor.py
+++ b/pyttb/ktensor.py
@@ -882,7 +882,7 @@ class ktensor(object):
                 vecs = []
                 for n in range(self.ndims):
                     vecs.append(self.factor_matrices[n][:, r])
-                res = res + self.weights[r] * other.ttv(np.array(vecs))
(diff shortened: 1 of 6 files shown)


brian-kelley:
I just figured out how to run pytest and there are a few issues, so consider this a WIP until I get those fixed.

jdtuck:
Looks great

jdtuck:
@dmdunla I have ran into this on other networks, would be good to get this one approved

Issue #41, opened 2022-07-18
tensor.__getitem__ error slicing on more than one dimension


dmdunla:
```
>>> shape = (3, 3, 3, 3)
>>> data = np.arange(1, 82)
>>> tensorInstance = ttb.tensor().from_data(data, shape)
>>> T = ttb.tensor().from_data(data, shape)
>>> T[0,0,0,0]
1
>>> T[0,0,0,:]
tensor of shape 3
data[:] = 
[ 1 28 55]
```

```
>>> T[0,0,:,:]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "c:\users\dmdunla\dev\github.com\pyttb\pyttb\tensor.py", line 1305, in __getitem__
    a = ttb.tensor.from_data(np.transpose(newdata, np.concatenate((kpdims, rmdims))))
  File "<__array_function__ internals>", line 5, in transpose
  File "C:\Users\dmdunla\Anaconda3\lib\site-packages\numpy\core\fromnumeric.py", line 660, in transpose
    return _wrapfunc(a, 'transpose', axes)
  File "C:\Users\dmdunla\Anaconda3\lib\site-packages\numpy\core\fromnumeric.py", line 57, in _wrapfunc
    return bound(*args, **kwds)
ValueError: axes don't match array
```

PR #50, opened 2023-02-04
TENSOR: Fix slices ref when return value isn't scalar or vector. #41

ntjohnson1:
A proposed fix to: https://github.com/sandialabs/pyttb/issues/41

For general tensors we should be able to defer to numpy slicing mechanics as the easiest solution. There is some ambiguity when slicing across multiple dimensions when it comes to retaining or dropping singleton dimensions. I don't have MATLAB access so I am not sure how it is handled there. All of our tests still pass and we never specifically enforced this so it might not be an issue.
Example:
```
import numpy
tensor = np.ones((3,3,3,3))
sliced = tensor[:, 0, :, 0]
sliced.shape # in numpy this is a matrix so shape is (3,3) but one could argue (1, 3, 1, 3) is possibly useful/valid
```

This came up when validating TTENSOR implementation so is a pre-requisite for that PR.

### DIFF
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -1272,7 +1272,6 @@ class tensor(object):
             rmdims = []   # dimensions to remove
 
             # Determine the new size and what dimensions to keep
-            # Determine the new size and what dimensions to keep
             for i in range(0, len(region)):
                 if isinstance(region[i], slice):
                     newsiz.append(self.shape[i])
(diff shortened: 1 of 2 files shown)


dmdunla:
@ntjohnson1 With the latest merged PR, your forked version for this PR is now behind the main branch and so the regression tests are failing. Can you update your forked version so the tests pass?

ntjohnson1:
> @ntjohnson1 With the latest merged PR, your forked version for this PR is now behind the main branch and so the regression tests are failing. Can you update your forked version so the tests pass?

I rebased to top of main. It looks like now that I'm approved my force push auto-triggered tests. I'll hold off on Tucker ALS and formatting until Tucker lands and we discuss the formatting changes since that one increases test time/has more opinionated components.

dmdunla:
The new behavior is the same at the current TTB behavior in MATLAB, which contracts singleton dimensions after slicing.

Python:
```
>>> T = ttb.tensor.from_data(np.ones((3,3,3,3)))
>>> T[:,:,:,0]
tensor of shape 3 x 3 x 3
data[0, :, :] = 
[[1. 1. 1.]
 [1. 1. 1.]
 [1. 1. 1.]]
data[1, :, :] = 
[[1. 1. 1.]
 [1. 1. 1.]
 [1. 1. 1.]]
data[2, :, :] = 
[[1. 1. 1.]
 [1. 1. 1.]
 [1. 1. 1.]]
```

MATLAB:
```
>> T = tensor(@ones,[3,3,3,3]);
>> T(:,:,:,1)
ans is a tensor of size 3 x 3 x 3
	ans(:,:,1) = 
	     1     1     1
	     1     1     1
	     1     1     1
	ans(:,:,2) = 
	     1     1     1
	     1     1     1
	     1     1     1
	ans(:,:,3) = 
	     1     1     1
	     1     1     1
	     1     1     1
```

PR #53, opened 2023-02-04
Tucker als

ntjohnson1:
This is a single commit on top of https://github.com/sandialabs/pyttb/pull/51. Mostly was using it as a sanity check of ttensor while I waited on approval to put up PR. Surprisingly required no changes to ttensor.

Two notes listed as TODO in the PR:
- If `normresidual` is supposed to have an abs inside the square root (the Tucker_als papers are behind papers and I didn't dig that hard. Not immediately obvious that normX should always be larger) then we should probably push that to MATLAB tensortoolbox as well
- `tensor.ttm` has an ambiguity in python because there is no negative zero. I special cased for this PR since we only use it one location. However `tensor.ttm(dims=None, exclude_dims=None)` is probably more explicit and avoids this subtle bug where you get only the first index instead of all but the first when using `-n`


dmdunla:
@ntjohnson1 Can you provide an example where the norm of the tensor squared is less than the norm of the core squared? I cannot reproduce this in MATLAB.

ntjohnson1:
@dmdunla this is from the test_tucker_als file
```
import numpy as np
import pyttb as ttb
data = np.array([[29, 39.], [63., 85.]])
shape = (2, 2)
params = {'data': data, 'shape': shape}
tensorInstance = ttb.tensor().from_data(data, shape)
(Solution, Uinit, output) = ttb.tucker_als(tensorInstance, 2)
```

With the absolute value I get something like:
```
 NormX: 116.43023662262307 Core norm: 129.9474584138604
 Iter 0: fit = 5.043460e-01 fitdelta = 5.0e-01
 NormX: 116.43023662262307 Core norm: 116.43023662262307
 Iter 1: fit = 1.000000e+00 fitdelta = 5.0e-01
 NormX: 116.43023662262307 Core norm: 116.43023662262307
 Iter 2: fit = 1.000000e+00 fitdelta = 0.0e+00
```

When I remove the absolute value I get nans for fit and fitdelta.

It's certainly possible there is a bug in my implementation so we could pick an example that converges in MATLAB and do a comparison. Or could try this one in MATLAB to see. I did get access to the Tucker references and based on my understanding there isn't a hard requirement for normX to always be greater than the core norm.

ntjohnson1:
@dmdunla this was bothering me so I got my hands on the MATLAB free trial. There is a small but in ttm for ttensor that I will put up a fix for in a minute. It's related to zero vs one indexing as pointed out in this issue already actually https://github.com/sandialabs/pyttb/issues/62

However, this absolute value issue is relevant but not a bug in the MATLAB tensortoolbox but is hidden based on the formatted printing:
```
>>> data = [29.,39.;63.,85.];
a_tensor = tensor(data);
Uinit = cell(2,1);
Uinit{2} = [0.05511886, 0.31326444; 0.28527485, 0.41035856];
result = tucker_als(a_tensor, 2, 'init', Uinit);
normx = norm(a_tensor);
normcore = norm(result.core);
fit = 1 - (sqrt(normx^2 - normcore^2)/normx);
fprintf('fit = %e', fit)
```
Results in:
fit = 1.000000e+00
However if we print the variable itself it is complex
```
fit
```
fit =

                          1 -   1.6381901198008e-08i
This doesn't seem to cause issues for MATLAB because apparently if we compare fit to a real it only considers the real portion so fit > 0 is true and fit < 2 is also true.

Issue #62, opened 2023-02-25
Address ttm dims=0 ambiguity


ntjohnson1:
Identified here https://github.com/sandialabs/pyttb/pull/53 MATLAB is 1 indexed so -index represents all indices EXCEPT for the one listed. However since python is 0 index -0 is the same as 0 so there is an ambiguity. One proposal is `tensor.ttm(dims=None, exclude_dims=None)`. Would need to see if this ambiguity is the same throughout other tensor implementations and should be updated uniformly/checked in algorithms that use the tensors.

ntjohnson1:
It looks like tt_dimscheck is also implicitly 1 indexed for our negative dims so we probably need to actually update this interface deeper where negative indices are used.

dmdunla:
> Identified here #53 MATLAB is 1 indexed so -index represents all indices EXCEPT for the one listed. However since python is 0 index -0 is the same as 0 so there is an ambiguity. One proposal is `tensor.ttm(dims=None, exclude_dims=None)`. Would need to see if this ambiguity is the same throughout other tensor implementations and should be updated uniformly/checked in algorithms that use the tensors.

I like the proposed change to use `tensor.ttm(dims=None, exclude_dims=None)`

This will require changes to (at least)
- [ ] `pyttb_utils.tt_dimscheck`
- [ ] `sptensor.ttm`
- [ ] `tensor.ttm`
- [ ] `ttensor.ttm`

ntjohnson1:
PR merged that resolved tt_dimscheck, as well as its usage in ttm, ttv, and ttsv

PR #66, opened 2023-03-11
TUCKER_ALS: TTM with negative values is broken in ttensor (#62)

ntjohnson1:
* Replace usage in tucker_als
* Update test for tucker_als to ensure result matches expectation
* Add early error handling in ttensor ttm for negative dims

### DIFF
--- a/pyttb/pyttb_utils.py
+++ b/pyttb/pyttb_utils.py
@@ -198,7 +198,7 @@ def tt_dimscheck(
 
     # Fix "minus" case
     if np.max(dims) < 0:
-        # Check that all memebers in range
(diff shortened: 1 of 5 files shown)


ntjohnson1:
See initial tucker_als PR https://github.com/sandialabs/pyttb/pull/53 for more details. My initial test just checked the fit value however because of the negative indexing we were effectively fitting a totally different tensor. Compared to MATLAB and things match now for the trivial small example and added a self consistency check to make sure our solution matches the tensor we are trying to factor.

dmdunla (approved):
Looks good. We'll have to add an Issue around include/exclude dims.

Issue #63, opened 2023-02-25
Add pylint and typing across pyttb


ntjohnson1:
This will probably need to be split across multiple sub-tasks/prs. The proposal and initial support landed here https://github.com/sandialabs/pyttb/pull/54. They are not automatically enforced yet

ntjohnson1:
I think for improving readability and extensibility starting with the tensor implementations is the way to go.
For that scoping I will track to this initial layout. For most of the files I'll type and lint in one go.

## Linting
- [x] tensor https://github.com/sandialabs/pyttb/pull/54
- [ ] sptensor (WIP)
- [ ] ktensor
- [ ] ttensor
- [ ] pyttb_utils (WIP)

## Typing
- [x] tensor https://github.com/sandialabs/pyttb/pull/54
- [ ] sptensor (WIP)
- [ ] ktensor
- [ ] ttensor
- [ ] pyttb_utils (This will probably wait until the tensor typing is complete)

dmdunla:
Is the linting working on the `main` branch at this point? When I tried the following using python 3.9.16, pytest 7.1.2, pylint 2.9.6, and mypy 0.981 on Windows:

```
pytest . --packaging
```
I got the following error:
```
================================================================ FAILURES ================================================================ 
______________________________________________________________ test_linting ______________________________________________________________ 

    @pytest.mark.packaging
    def test_linting():
        """Confirm linting of the project is enforce"""

        enforced_files = [
            os.path.join(os.path.dirname(ttb.__file__), f"{ttb.tensor.__name__}.py"),
        ]
        # TODO pylint fails to import pyttb in tests
        # add mypy check
        root_dir = os.path.dirname(os.path.dirname(__file__))
        toml_file = os.path.join(root_dir, "pyproject.toml")
        for a_file in enforced_files:
>           subprocess.run(f"pylint {a_file} --rcfile {toml_file}", check=True)

tests\test_package.py:32:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _  

input = None, capture_output = False, timeout = None, check = True
[...]

PR #77, opened 2023-03-18
Lint pyttb_utils and lint/type sptensor

ntjohnson1:
Related to https://github.com/sandialabs/pyttb/issues/63
The commits mostly break down the work but this covers pyttb_utils and sptensor for pylint, but just sptensor for typing. I think will need the tensor implementations to be typed for the utils that act on tensors to place nice with types.

Continuing to add typing across the base classes. The native python numeric types vs numpy numeric types isn't generally resolved yet. I think the typing is helping to see all the places we use it but I probably need to make an issue to track this globally. I've moved towards `isinstance(variable, (int,float,np.generic))` to basically catch numpy scalars but I'd need to dig around a bit more to see if there is a standard for switching back and forth while also playing nicely with typing.

I mentioned before that our pyttb_utils today are both used by the tensor types internally but also act on tensor types. It might be clearer to split those out. I moved the pyttb_utils that ONLY sptensor uses internally into the sptensor file. I think after some other changes this wasn't strictly necessary, so can move them back if preferred.

### DIFF
--- a/pyttb/pyttb_utils.py
+++ b/pyttb/pyttb_utils.py
@@ -1,19 +1,20 @@
 # Copyright 2022 National Technology & Engineering Solutions of Sandia,
 # LLC (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the
 # U.S. Government retains certain rights in this software.
-
+"""PYTTB shared utilities across tensor types"""
 from inspect import signature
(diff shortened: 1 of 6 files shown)


dmdunla (approved):
Looking good!

PR #99, opened 2023-04-26
Remove warning for nvecs:

ntjohnson1:
I think an enhancement that probably isn't blocking for 2.0 is considering a more standard logging control. Some algorithms take log levels that aren't consistent with each other and use prints. I've been move more things over to `logging` we can probably add a `pyttb` level logging control rather than the root logger for a file. This allows a more standard way to control verbosity.

All of that to say for now I moved the warning to a debug level log, because I think we added the warning initially in case someone hit some weird numeric behavior with the cast (I don't remember entirely).

### DIFF
--- a/pyttb/ktensor.py
+++ b/pyttb/ktensor.py
@@ -5,6 +5,7 @@
 """Classes and functions for working with Kruskal tensors."""
 from __future__ import annotations
 
+import logging
 import warnings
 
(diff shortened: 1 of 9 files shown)


dmdunla:
This was a carryover from Matlab. I did not find it useful there, either.

Issue #105, opened 2023-05-04
Rename Contributor Guide to Contributing


ntjohnson1:
Github has some magic around detecting contributor guides specified [here](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) and if we name the file appropriately it should show details at https://github.com/sandialabs/pyttb/contribute instead of the blank page we currently have.

PR #106, opened 2023-05-04
Rename contributor guide for github magic

ntjohnson1:
Renames the file since I didn't know about the github naming convention when I added it.

Non-functional change with details in the issue: https://github.com/sandialabs/pyttb/issues/105


dmdunla:
I think the `README.md` needs to change to point to this file. Will that happen automagically when the file name changes?

ntjohnson1:
Yes. I forgot to do a grep. Just did one and updated README.

Issue #108, opened 2023-05-22
Clean Up Tensor/Sptensor setitem (and getitem to match)


ntjohnson1:
This is an extension from the long discussion here but the short summary is:
This is the example from MATLAB https://github.com/sandialabs/pyttb/issues/92#issuecomment-1537171944 with this being the example from Danny with additional elaboration on expected behavior https://github.com/sandialabs/pyttb/issues/92#issuecomment-1540846022. Will follow up on more details about the expected python interface to confirm. I think the pieces we have are fairly close already so this shouldn't impact correctness of existing algorithms but rather to confirm this as a toolbox acts as expected.

ntjohnson1:
Closed with https://github.com/sandialabs/pyttb/pull/116

dmdunla:
Re-opening issue, as the behavior described in https://github.com/sandialabs/pyttb/issues/92#issuecomment-1540846022 seems to work for `__setitem__` but not for `__getitem__`. Here are the cases for `tensor` that MATLAB supports (with examples illustrating my understanding of what the corresponding Python call should look like for a 4-way tensor). `sptensor` may need to be addressed as well, but this is a start for `tensor`.

Intended behavior (to align with MATLAB):
1a. Single subscript -> scalar: `X[0,0,0,0]`
1b. Slice -> tensor: `X[0,0,0,:]`, `X[:,0,0,:]`, or `X[0:2,[1,3],0,:]`
2a. Array (p x N) of subscripts -> array (p x 1): `X[np.array([[0,0,0,0],[2,3,1,0]])]`
2.b Array (p x 1) of linear indices -> array (p x 1): `X[np.array([[0,5,6,22]]).T]`

Current behavior:
1a. Working as in MATLAB.
1b. Working as in MATLAB.
2a. Does not seem to be working.  Restricted to 2 x N array input and returns a slice not (p x 1) array.
2b. Does not seem to be working. Does not accept array input (only list) and return a 1-D array (which may be OK since we are using numpy, but this breaks from MATLAB).

Steps to reproduce error:

MATLAB:
```matlab
>> X = tensor(1:24,[3 4 2 1])
X is a tensor of size 3 x 4 x 2 x 1
	X(:,:,1,1) = 
	     1     4     7    10
	     2     5     8    11
	     3     6     9    12
	X(:,:,2,1) = 
	    13    16    19    22
	    14    17    20    23
	    15    18    21    24

>> % Case 1a
>> X(1,1,1,1) 
ans =
[...]

PR #116, opened 2023-05-28
Tensor indexing

ntjohnson1:
There is definitely cleanup that could be done around our indexing tests and potentially the logic. There are a lot of different paths through the code based on a variety of conditions that makes it complicated to reason about.

This proposes a solution to: https://github.com/sandialabs/pyttb/issues/108

Here are all the different combinations that I could think of and expected behavior based on Danny's description in the linked issue/discussion. I basically used this as my burn down list to confirm consistency. If there are disagreements here I think the small examples help for concreteness (and I baked some into our unit tests). The fundamental problem was the sub-scripts had different meanings in pyttb between sptensor and tensor. Ultimately, that made me question how everything interacted so I just explored the intuitiveness of a variety of access patterns that seemed reasonable. One note, do we even need the `extract` keyword? It doesn't seem to work as described in matlab (unless I broke my local version).
```matlab
 >> X = tensor(rand(3,4,2,1),[3 4 2 1]);
 >> a = X([1:6]');
 >> b = X([1:6]', 'extract');
 >> class(a)
ans =
    'double'
>> class(b)
ans =
    'double'
```

Big decision tree covering functionality that this PR now supports
### Linear
1. Single index
```python
        from pyttb import tenones
        b = tenones((3,3))
        b[0] = 9
        assert b[0] == 9
```
2. Multiple linear indices
```python
        from pyttb import tenones
        b[[0,1]] = 2
        assert b[0] == 2
        assert b[1] == 2
        assert np.array_equal(b[[0,1]], [2,2])
```
### Sub-tensor:
1. Singular indices
```python
        from pyttb import tenones
        b = tenones((3,3))
        b[0,0] = 9
        assert b[0,0] = 9
        #(with growth)
        b[0, 0, 0] = 11
        assert b[0,0,0] == 1
```
2. With slices
```python
        from pyttb import tenones
        b = tenones((3,3))
        b[:, 0] = 11
        assert b[0,0] == 11
        assert b[1,0] == 11
        assert b[2,0] == 11
        assert np.all(b[:,0].data == 11)
```
3. With multi-indices (not range since python has slices)
```python
        from pyttb import tenones
        b = tenones((3,3))
        b[[0,1],[0,1]] = 11
        assert b[0,0] == 11
        assert b[1,1] == 11
        assert np.all(b[[0,1],[0,1]].data == 11)
```
### Sub-scripts:
1. Only one option of being explicit
```python
        b = tenones((3,3))
        b[np.array([[1,1],[0,2]])] = 2
        assert np.all(b[np.array([[1,1],[0,2]])] == 2)
```
### Sub-tensor:
1. Singular indices
```python
        from pyttb import sptensor
        a = sptensor()
        a[0,0] = 1
        assert a[0,0] == 1
```
2. Slices
```python
        from pyttb import sptensor
        a = sptensor()
        a[2,2] = 1
        a[:, 2] = 2
        assert a[0, 0] == 2
        assert a[1, 0] == 2
        assert a[2, 0] = 2
        assert np.all(a[:, 2].vals == 2)
```        
3. Multi-indices
```python
        from pyttb import sptensor
        a = sptensor()
        a[[0,1], 0] = 1
        assert a[0,0] == 1
        assert a[1,0] == 1
        assert np.all(a[[0,1], 0].vals == 1)
```
### Sub-scripts:
1. Only one option of being explicit
```python
        a = sptensor()
        a[np.array([[1,1],[0,2]])] = 2
        np.array_equal(a[np.array([[1,1],[0,2]])], np.array([[2]]*2))
```

### DIFF
--- a/pyttb/sptensor.py
+++ b/pyttb/sptensor.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 import logging
 import warnings
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 from typing import Any, Callable, List, Optional, Tuple, Union, cast, overload
 
 import numpy as np
@@ -620,7 +620,7 @@ class sptensor:
             if self.shape != other.shape:
                 assert False, "Sptensor and tensor must be same shape for innerproduct"
             [subsSelf, valsSelf] = self.find()
-            valsOther = other[subsSelf, "extract"]
+            valsOther = other[subsSelf.transpose(), "extract"]
             return valsOther.transpose().dot(valsSelf)
 
         if isinstance(other, (ttb.ktensor, ttb.ttensor)):  # pragma: no cover
@@ -685,7 +685,7 @@ class sptensor:
 
         if isinstance(B, ttb.tensor):
             BB = sptensor.from_data(
-                self.subs, B[self.subs, "extract"][:, None], self.shape
+                self.subs, B[self.subs.transpose(), "extract"][:, None], self.shape
             )
             C = self.logical_and(BB)
             return C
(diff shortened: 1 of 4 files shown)


dmdunla:
`extract` does not seem to work as intended in the MATLAB code either. Will create issue in that code.

dmdunla (approved):

Issue #14, opened 2022-06-28
Testing: implement tests for full coverage


dmdunla:
If possible, implement tests to provide full coverage of TensorToolbox code:

```
pytest --cov=pyttb  tests/ --cov-report=term-missing
```

```
Name                   Stmts   Miss  Cover   Missing
----------------------------------------------------
pyttb\__init__.py         23      1    96%   31
pyttb\cp_als.py           87      0   100%
pyttb\cp_apr.py          567     78    86%   84, 86, 88, 110, 197-198, 229, 247-249, 342, 351, 374-384, 419-432, 464-469, 495, 536, 540-541, 649, 658, 681-691, 719-732, 781-786, 808, 813, 819, 871-872, 1016, 1034-1037, 1044-1045, 1230, 1297-1299, 1343
pyttb\export_data.py      63      1    98%   54
pyttb\import_data.py      60      4    93%   14, 24, 61, 72
pyttb\khatrirao.py        22      0   100%
pyttb\ktensor.py         468      5    99%   772, 812-816
pyttb\pyttb_utils.py     237      2    99%   293, 328
pyttb\sptenmat.py          4      0   100%
pyttb\sptensor.py        914      4    99%   117, 562, 608, 637
pyttb\sptensor3.py         4      0   100%
pyttb\sumtensor.py         4      0   100%
pyttb\symktensor.py        4      0   100%
pyttb\symtensor.py         4      0   100%
pyttb\tenmat.py          182      1    99%   183
pyttb\tensor.py          557      3    99%   1105, 1219, 1394
pyttb\ttensor.py           4      0   100%
----------------------------------------------------
TOTAL                   3204     99    97%
```

dmdunla:
All but `cp_apr` has 100% coverage using `pytest` and `coverage`. Coveralls must use some other process to compute coverage, as the published coverage there is different than this:

```
Name                   Stmts   Miss  Cover   Missing
----------------------------------------------------
pyttb/__init__.py         25      0   100%
pyttb/cp_als.py           87      0   100%
pyttb/cp_apr.py          569     18    97%   273-274, 635, 821, 898-899, 998, 1006, 1012, 1255, 1275-1278, 1292-1293, 1574-1576
pyttb/export_data.py      68      0   100%
pyttb/hosvd.py            58      0   100%
pyttb/import_data.py      64      0   100%
pyttb/khatrirao.py        22      0   100%
pyttb/ktensor.py         519      0   100%
[...]

PR #128, opened 2023-06-03
14 testing implement tests for full coverage

dmdunla:
Good for now. Some of the remaining uncovered lines in `cp_apr` are due to edge cases that are difficult to reproduce. No tests in MATLAB cover these either.


dmdunla:
Closes #14

PR #132, opened 2023-06-05
Cleanup regressions

ntjohnson1:
Test if we can resolve this https://github.com/sandialabs/pyttb/pull/131#issuecomment-1577529240

Revert some changes, enforce pylint, and fix a bug because windows path representation is silly.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -41,7 +41,7 @@ jobs:
 #        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: Run tests
       run: |
-        coverage run --source pyttb -m pytest tests/
+        coverage run --source pyttb -m pytest tests/ --packaging
         coverage report
     - name: Upload coverage to Coveralls
       uses: coverallsapp/github-action@v2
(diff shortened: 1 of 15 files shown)


ntjohnson1:
Is there a reason we moved away from just calling coveralls directly? That coveralls github action has an issue open specifically related to it having issues with python. They point to this python specific package to support https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support

https://github.com/lemurheavy/coveralls-public/issues/1610#issuecomment-1020451310

dmdunla (approved):

dmdunla:
Here's an example of what's happening in coveralls when including the file docstring at the top; the class docstring appears as uncovered "code". If there is a problem with coveralls, we could consider moving to another code coverage tool; I am not too fluent in what's available.

dmdunla:
Here's the example:

https://coveralls.io/builds/60539470/source?filename=pyttb%2Fktensor.py

ntjohnson1:
If I understand correctly it's our GitHub action integration with coveralls rather than coveralls itself. Feel free to assign an issue to me to investigate or I'll file one later this week.

dmdunla:
Here's the output of running coveralls directly in a clean conda env locally with these updates:

https://coveralls.io/jobs/122739175

This was run using the following (per coveralls docs):
```
pip install coveralls
coverage run --source pyttb -m pytest tests/ --packaging
COVERALLS_REPO_TOKEN=<removed> coveralls
```

dmdunla:
I could not get this to run with this level of coverage unless I created a new conda env. So, I think we may want to be more strict about our minimum versions of dependencies that we use for development.

PR #133, opened 2023-06-06
Coveralls: Try updating to default suggestion

ntjohnson1:
Following the coveralls python package suggestion here seems to resolve our issues with weird coverage in our github action. https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support

NOTE: Do we use the sphinx_rtd_theme anywhere for our current testing/deployment? I don't think so. If we do we should probably update our automation to `pip install .[dev,doc]` instead of manually specifiying the theme (manually specifying coveralls since that is only for the action makes sense still). But that's more energy than I have for this tonight.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install coverage coveralls sphinx_rtd_theme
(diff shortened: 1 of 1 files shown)


ntjohnson1:
If I understand correctly the issues we were seeing before had to do with the coveralls github action and package itself rather than any of our source code. I still think it is probably good practice for us to have a little more specificity on our pacakage dependencies but that doesn't appear to be the cause of the bug we saw with our coverage

dmdunla (approved):

PR #135, opened 2023-06-07
Placed the 'whos' function within pyttb.utils. 

DeepBlockDeepak:
This function simulates MATLAB's 'whos'. Tried to add the documentation appropriately but it may need a second look.

### DIFF
--- a/pyttb/pyttb_utils.py
+++ b/pyttb/pyttb_utils.py
@@ -3,6 +3,7 @@
 # LLC (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the
 # U.S. Government retains certain rights in this software.
 
+import sys
 from inspect import signature
 from typing import Optional, Tuple, overload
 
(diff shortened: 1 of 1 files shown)


DeepBlockDeepak:
Looks like I bit off more than I could chew with the documentation, which seems to be the source of the failed checks. I'm going to remove the faulty looking bits and retain them here for posterity:
```python
    Returns
    -------
    None
        Prints a summary of each variable to the console but does not return a value.

    Example
    -------
    >>> X = ttb.tensor().from_data(np.random.rand(5), shape=(5,))
    >>> Y = ttb.tensor().from_data(np.random.rand(4,3,1), (4,3,1))
    >>> whos('X', 'Y')
    Name      Size           Bytes     Class               Num. Attributes
    ---------------------------------------------------------------------
    X         5              56        tensor              36
    Y         4x3x1          56        tensor              36
```

I replaced that with : 
```python
    Returns
    -------

    """
    <beginning of code>
```

ntjohnson1:
A couple of thoughts (non-blocking if Danny disagrees and wants to add the function)

1. Since this is a MATLAB feature instead of a Tensor Toolbox feature I'd prefer to avoid us handling it unless it is critical somewhere. If I understand correctly this is just a nice to have for the docs. Also since our tutorials are in Jupiter notebooks can't we just rely on IPython's who's feature? https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-who
2. We use doctest to check the documentation correctness. https://docs.python.org/3/library/doctest.html you should be able to run that as a part of your tests locally if you wanted to debug what's going wrong with the example format. If you look at the bottom of ktensor there is the snippet to just run the doctest for a single file for easier iteration.

DeepBlockDeepak:
Thanks Nick. I'll definitely be looking at all those things. Also, turns out I made the mistake of working on this instead of making it into a discussion topic.

ntjohnson1:
I am not sure if we have a place to add philosophical statements that drive decisions, but I think this did highlight one we hadn't explicitly stated before. We want to capture the tensor functionality but not reimplement MATLAB functionality, which is part of the reason for pushback here. Otherwise the scope of the project might grow in weird undesired ways. That detail probably is helpful if something like this comes up in the future or onboarding new contributors.

Issue #6, opened 2022-06-08
Add support for different index bases in `import_data`


dmdunla:
Currently, `import_data` assumes that incoming data is 0-based index, but compatibility with Matlab Tensor Toolbox output requires that 1-based index be supported as well.

etphipp:
Actually that isn't correct.  Currently pyttb assumes the incoming data is 1-based.

etphipp:
Added PR #144 to address this.

PR #138, opened 2023-06-08
Enable integration with pygenten

etphipp:
The primary purpose of this PR is to enable integration of pygenten with pyttb.  The python package pygenten is a set of python bindings for the [GenTen](https://gitlab.com/tensors/genten) package and provides HPC-oriented implementations of several CP decomposition methods.  The changes here enable pygenten as a back-end implementation of (currently) CP-ALS, if pygenten is available.  It also adds a `copy` argument to several pyttb construction functions to allow conversions between pyttb and pygenten objects without copying the underlying data (defaulting to `True` to preserve prior behavior).  Other small changes include fixing an ordering bug when writing dense tensors to a file that I discovered in this work (numpy always writes files using "C" ordering, regardless of the ordering of the data in memory) and added an `index_base` argument  to `import_data()` to support sparse tensor files using 0-based indexing (such as those produced by GenTen).

### DIFF
--- a/pyttb/cp_als.py
+++ b/pyttb/cp_als.py
@@ -143,7 +143,7 @@ def cp_als(tensor, rank, stoptol=1e-4, maxiters=1000, dimorder=None,
     else:
         assert False, "The selected initialization method is not supported"
 
-    if isinstance(tensor, ttb.tensor) and genten_backend:
+    if genten_backend and (isinstance(tensor, ttb.tensor) or isinstance(tensor, ttb.sptensor)):
         # Call pygenten
         args = {}
         args['init'] = init
--- a/pyttb/cp_als.py
+++ b/pyttb/cp_als.py
@@ -146,29 +146,21 @@ def cp_als(tensor, rank, stoptol=1e-4, maxiters=1000, dimorder=None,
     if isinstance(tensor, ttb.tensor) and genten_backend:
         print("Genten is supported")
 
-        x = pygenten.Tensor(tensor.data)
-        u = pygenten.Ktensor(init.weights, init.factor_matrices);
-
+        # Call pygenten
         args = {}
-        args['init'] = u
+        args['init'] = init
         args['rank'] = rank
         args['maxiters'] = maxiters
         args['tol'] = stoptol
         args['printitn'] = printitn
-        u, perfInfo = pygenten.cp_als(x, **args)
+        M, perfInfo = pygenten.cp_als(tensor, **args)
 
+        # Set output
         output = {}
         output['params'] = (stoptol, maxiters, printitn, dimorder)
         output['iters'] = perfInfo.lastEntry().iteration
         output['normresidual'] = perfInfo.lastEntry().residual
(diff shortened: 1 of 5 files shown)


ntjohnson1:
Hey Eric! This looks cool. I didn't realize you were exposing Genten via pybind. I have a few thoughts that hopefully aren't too cumbersome. My vote would be to break this into 3 more focused PRs (or could do 2).

1. Arbitrary index base for import data (and the export data order) which we already have a ticket for https://github.com/sandialabs/pyttb/issues/6
   a. We also would like some testing to validate the changes in the contributing doc it provides the commands to create the expected environment and command to run our code coverage test to ensure we cover the changes (our tests should run locally quite quickly). Also would love feedback on any confusion or issues with following the contributing guide.
   
1. Adding the `copy` option (again with tests) could you file a ticket for this? Reducing copies in pyttb has been in the back of my mind for optimization anyway so this seems like a reasonable first step

1. The actual pygenten usage. I'd like to have more discussion on a ticket about this. I think supporting additional backends is definitely interesting, but probably needs a little bit more thought about how to do it cleanly. For example, pygenten doesn't look like it is on pypi so we probably need to figure out how to instruct someone to use that backend and how to test our integration works.

There are also a few conflicts with main to resolve.

tgkolda:
+1 for being about to use genten in pyttb

dmdunla:
Following up on comments/suggestions from @ntjohnson1 :
1. This should be handled as a separate PR.
  a. Agreed about testing, please provide updates to [`test_import_export_data.py`](https://github.com/sandialabs/pyttb/blob/main/tests/test_import_export_data.py) and follow the guidance for contributing in [`CONTRIBUTING.md`](https://github.com/sandialabs/pyttb/blob/main/CONTRIBUTING.md), specifically regarding pr-commit checks for formatting (`isort`, `black`)
2. I think this identifies an issue with the current approach. The example of `ktensor.from_data` making copies of the factor matrices but not the weights, coupled with `ktensor.from_tensor_type` maink copies of both illustrates the problem. I'll create an issue so that we discuss what we want in the code moving forward and can handle all classes consistently. This should be handled as a separate PR.
3. Agreed about using `pygenten` without a path for support. @etphipp what is the plan for `pygenten` dev/deployment/maintenance? Is the goal to deploy via PyPI or some other manager? This would be helpful for our support of it. I understand it is optionally imported here, but I would like to make sure we can test in the `pyttb` CI tests as needed.

etphipp:
Working on it.  Sorry about the conflicts, non-compliant formatting, and missing tests.  I thought I had done all the steps for contributing, but forgot my branch was based on a very out-of-date fork.  So far I have PR #143 for the transpose issue.

etphipp:
Also added PR #144 for the import_data contribution.

ntjohnson1:
> Sorry about the conflicts, non-compliant formatting, and missing tests.

No problem! We are still working out the details. Thanks for the fixes and the quick turn around. I am going to go ahead and close this PR since we have a path forward.

Issue #142, opened 2023-06-08
tensor.export_data() outputs the transpose of the tensor


etphipp:
According to the numpy [docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tofile.html#numpy.ndarray.tofile), `numpy.tofile()` always writes out the array with 'C' ordering, regardless of the ordering of the array.  This means dense tensors written to files by pyttb are the transpose of the tensor, since they are assumed to use 'F' ordering.

PR #143, opened 2023-06-08
Fix dense tensor export to write the tensor using the assumed 'F' ordering

etphipp:
This PR resolves issue #142 by
* changing the dense tensor export tests to exhibit the bug in #142 by changing the tensor values to not all be identical
* fixes the bug by transposing the tensor before writing it


ntjohnson1 (approved):
This looks reasonable to me, but I am not super familiar with our import/export code paths.

PR #144, opened 2023-06-09
Add support for different index bases in import_data

etphipp:
Addresses issue #6 by adding an `index_base` optional argument to `import_data()` (which defaults to 1 to preserve current behavior).  Also added a test that reads in a 0-based sptensor.


dmdunla:
@etphipp can you request a review from me when you are ready for this (or any) PR to be merged? (We may need to add this step to the CONTRIBUTING.md guide.)

ntjohnson1 (approved):

dmdunla (approved):

Issue #107, opened 2023-05-06
Explore Adding Tutorials to RTD


ntjohnson1:
I know I haven't been running the tutorials. We could add some tests to confirm the notebooks run. But it might be nicer to include them as a part of our read the docs release. https://docs.readthedocs.io/en/stable/guides/jupyter.html

dmdunla:
Let’s assign this to our new team member when they start in late May

PR #163, opened 2023-06-30
Adding Tutorials to ReadTheDocs

DeepBlockDeepak:
Created tutorials.rst, while updating conf.py and index.rst. (#107). 

Changes here:

1. Use myst  to build our notebooks into read the docs
   1. Alternative of nbsphinx has an external pandoc dependency which can't just be managed by pip. Much more painful.
1. Remove notebook `contents` block since jupyterlab auto creates a table of contents on the left AND the RTD build does the same. (We should probably do that for the other staged options as well. Just drive things via jupyter header levels)
1. Move notebooks to /docs/source/tutorial. We probably will want to include some details on that in our README or somewhere else for people that want to run notebooks themselves.
   1. This is mostly related to sphinx limitations for pulling in our of source content

We don't have this automated for CI yet. So here is the rough repro (we might need to add an explicit jupyter-notebook dependency to our doc option, I assume you had to install it into your venv to run any of the notebooks good for us to clean that up too):

```bash
pip install -e ".[doc]"
sphinx-build ./docs/source ./docs/build
```
Then you should be able to open the built local read the docs to look at it.


docs/source/tutorials.rst
```diff
@@ -0,0 +1,9 @@
+Tutorials
+=========
+
+.. toctree::
+   :maxdepth: 2
+
+   ../../../tutorial/CP_APR
```
ntjohnson1, line 7: Did you test the sphinx build locally? It is broken with the out of tree build here for me. Blocking the PR for the moment. I will provide more details in a a little bit.

ntjohnson1 (changes requested):
This PR highlights the fact our current set of tests don't actually build or check our docs. There are github actions that will build a draft of the read the docs for easier review/validation the docs work. Depending on how hard those are it might be worthwhile to get that figured out to with all the doc changes we have staged. Alternatively, we can just build the docs locally off the PRs for review which isn't too cumbersome. I will probably open a branch to fix some of the issues here for Jordan to review/consider.

An additional conceptual point I didn't write in appropriate detail in the original issue is that of clearing our notebook output. If we look at all the tutorial PRs there is all kinds of meta data being added into git which will change every time a notebook is run which obscures diffs. In the link to the RTD approach it mentions some options. IMO the two broad options are:

1. Use an automated tool like nbdev, or a [commit filter](https://stackoverflow.com/questions/73218304/how-to-clear-jupyter-notebooks-output-and-metadata-when-using-git-commit)
1. Use a non-ipynb format that is more text based such as [jupytext](https://jupytext.readthedocs.io/)

All the notes above are for developer/reviewer experience. In practice the short term solution might be manual review, manually clear all our notebooks, and get this PR patched up and landed. Then we can revist smoothing out some of the process etc. That way the content is all in first (which is probably my preference).

ntjohnson1:
Ok I opened a [PR ](https://github.com/DeepBlockDeepak/pyttb/pull/1) into Jordan's branch for him to consider. If he is ok with that then we can loop Danny in on the final product to confirm he likes the direction. If Danny wants to weigh in earlier can feel free to take a look at my branch (it is this branch plus like two commits)

docs/source/tutorials.rst
DeepBlockDeepak: Whoops, looks like I definitely didn't test that out properly.

DeepBlockDeepak: I'm working on the previous Tutorial PR's where I'll be clearing out the Notebook outputs to help clean things up.

ntjohnson1: No problem! You are helping us to find processes we need to build, or describe better. Ideally, we make it challenging to break things (there are so many other things to keep track of, offloading as much as possible to tools or automation is usually the way to go).

DeepBlockDeepak:
Changes looked really good btw. Good call going with `myst_nb`.

ntjohnson1 (approved):

ntjohnson1:
I updated the description with the details for the PR for future reference. My brain dump above I can transfer to a ticket if everyone is on the same page for the direction we want to go for improving robustness. In the short term my preference is still to:
> manual review, manually clear all our notebooks, and get this PR patched up and landed

(as well as the other staged tutorial PRs)

To be explicit this is ready for review @dmdunla but I don't think is in a rush. It seems like Jordan and I were both able to make this work and both think it looks nice.

docs/source/tutorial/Tensors.ipynb
ntjohnson1, line 1: FYI Github wasn't smart enough on the merge commit to track the notebook rename on main to the move here. Will need to update the branch to show that we should move the new tensor notebook instead of generating a new one

dmdunla:
I have been using [`nbstripout`](https://pypi.org/project/nbstripout/) on another project and it has worked really well.

We can add it to the `dev` dependencies and use it proactively: `nbstripout tutorial/*.ipynb`

We can add it to the pre-commit runs.

We can add "install" it and store its use via `.gitattributes`.

dmdunla:
I'm going to work through the tutorial PRs first and then update this if needed before approving/merging.

ntjohnson1:
I opened a [PR](https://github.com/DeepBlockDeepak/pyttb/pull/3) into the branch to update this based on all the other tutorials that merged. So this should be good to go once that lands. I am fine with `nbstripout` as follow up here.

dmdunla:
@DeepBlockDeepak Waiting on the the new PR:
- https://github.com/DeepBlockDeepak/pyttb/pull/3

Issue #170, opened 2023-06-30
Sptensor indexing not working as expected


DeepBlockDeepak:
### Extracting Nested Item 

Extracting a nested tensor element from a 2D array input isn't working as expected. I'm not sure how to get extraction to work with a 2D(?) array as input, like in the MATLAB version. 

Ex:
```python
>>> import pyttb as ttb
>>> import numpy as np
>>> X = ttb.sptensor.from_data(np.array([[3,3,3],[1,1,0],[1,2,1]]), np.array([[3],[5],[1]]), (4, 4, 4)) # Create a sptensor.
>>> X[1,1,1] # a numpy.ndarray is extracted, NOT a single integer element, as in the MATLAB below.
array([[0]])
```

However, in [MATLAB ttb](https://tensortoolbox.org/sptensor_doc.html#28:~:text=X(1%2C2%2C1),ans%20%3D%0A%0A%20%20%20%20%200) it looks like:
```matlab
X(1,2,1) %<-- Extract the (1,2,1) element, which is zero.
ans =

     0
```

DeepBlockDeepak:
Checkpoint issue #171 will continue this work.

Issue #161, opened 2023-06-29
Determine if `end()` is necessary in pyttb classes


dmdunla:
Do we the `end()` methods in pyttb classes?

The `end()` method is TTB for MATLAB is used to identify the index of the last element of a tensor using linear indexing or the last index along a dimension when using subscript indexing. 

`numpy` , which is used for most data objects in pyttb classes, supports negative indexing but in a different way than what `end()` does.

Should we implement `end()` methods that wrap `numpy` negative indexing (e.g., `X[-1,-1,-1]` for subscript indexing or `X.flat[-1]` for linear indexing? Or remove the `end()` methods?

ntjohnson1:
I think we should avoid using `end()` for all indexing operations since that's very unnatural for python. I believe our slicing support already handles negative values. I was just looking at an example from sptensor `X[110:, 110:, 115:].isequal(X[-10:, -10:, -5:])` from this PR https://github.com/sandialabs/pyttb/pull/152 (after I apply some fixes to the sptensor indexing but nothing specific to negative indices).

tgkolda:
It looks like -1 should work fine in lieu of end. I would suggest to have some explanation of this in the documentation, especially for those transitioning from MATLAB.

dmdunla:
Answered this question. Will use -1 indexing in lieu of `end()` in class methods.

Follow up:
- #179

Issue #160, opened 2023-06-29
tensor/sptensor: `__getitem__()` determine use of `extract` parameter 


dmdunla:
The following, 
- `pyttb.tensor.__getitem__`
- `pyttb.sptensor.__getitem__`

can take a parameter, `extract`, as input in the corresponding TTB for MATLAB.

I suggest we remove this functionality, as it has caused several challenges in the implementation. Moreover, I believe that a second step of converting the output of `__getitem__` to a vector when needed would suffice for the intended use case.

[TTB MATLAB Isssue 73](https://gitlab.com/tensors/tensor_toolbox/-/issues/73) plans to update the docs to reflect the intended use there: for 1-way tensor/sptensor objects, the `extract` is used to request that the output should be a vector not a tensor/subtensor.

ntjohnson1:
I support dropping extract. It looks like we never really supported it in sptensor anyway.

DeepBlockDeepak:
Checkpoint issue #171 will continue this work.

PR #174, opened 2023-07-02
Sptensor tutorial indexing

ntjohnson1:
This PR is all scope creep. I've been battling indexing for a while and had been meaning to do a larger scale refactor to make things more obvious. I finally had some bandwidth for this but I think there is more to do related to simplifying and unifying code in setitem for sptensor and tensor. That will be another day.

As far as I am aware this should resolve:

- https://github.com/sandialabs/pyttb/issues/170 added tests for these but didn't touch the notebook since still in PR
   - NOTE: @DeepBlockDeepak I updated `sptensor()` to take a shape argument to simplify generating empty sptensors with starting shape
- https://github.com/sandialabs/pyttb/issues/160 pruned usage of `extract` from tensor and sptensor since it aided in the test cleanup
- https://github.com/sandialabs/pyttb/issues/161 kept `end()` but updated sptensor and tensor indexing to accept negative values for non-subscript options (subscripts are most explicit so need to be correct).

I broke out the different pieces by commit so it might be easier to review that way for motivation of individual changes then look at the resulting product as a whole to decide if there is anything else we'd like to change. In my opinion grouping the tests into test classes with smaller simpler test methods makes it clearer WHY we have certain indexing tests and where updated tests should go if we find more edge cases.

Some of the work here overlaps pretty strongly with the effort to add typing over all of pyttb. There are quite a few tests in our unit tests to exercise a bunch of different type combinations that I believe mypy should be able to statically validate/point out issues with as well continue our coverage.


DeepBlockDeepak:
Your updates work nicely with the sptensor tutorial's TODOS!

dmdunla:
I'll submit an issue to discuss the example in `test_sptensor.py/test_subtensor_assign_oddity`

Issue #61, opened 2023-02-25
Support general cores in TTENSOR


ntjohnson1:
[This ](https://github.com/sandialabs/pyttb/pull/51) added ttensor support for dense cores. MATLAB supports any core that follows an interface. We can probably use protocols and mypy to sanity check this. Alternatively may consider creating a base class for some shared pyttb tensor functionality.

DeepBlockDeepak:
Leaving note here to highlight issue #147 's `Core Type Conversion` note, and Nick's response to it. The Ktensor.ipynb will need to be updated accordingly, once this this `core` issue is resolved.

ntjohnson1:
MATLAB states generic cores are supported but special cases tensors and sptensor. For the scope of this ticket only adding support to sptensor.

PR #177, opened 2023-07-03
TTENSOR: Half-hearted support for sparse cores and factors

ntjohnson1:
My understanding is that there isn't a ton of desire to handle sparse TTENSORS. This is mostly just added to handle the tutorial. If it starts getting a lot of traffic we should probably revisit. As I mentioned in https://github.com/sandialabs/pyttb/issues/61 this just adds support of sptensor since it doesn't appear like MATLAB actually supports fully general cores.

Should probably get confirmation from @DeepBlockDeepak that the tutorial does in fact work with this. I have fairly high confidence.


DeepBlockDeepak:
Instancing a `ttensor` with a `sptensor` core is now working perfectly as far as I can tell, @ntjohnson1!

ntjohnson1:
> Instancing a `ttensor` with a `sptensor` core is now working perfectly as far as I can tell, @ntjohnson1!

Great! I've queued a bunch of stuff for Danny to look at. I think once all the initial drafts of the notebooks, and my bug fixes land you can probably do a single PR to capture all the updates to finalize them. Or whatever is easiest for you and Danny.

PR #178, opened 2023-07-03
Add RTD preview

ntjohnson1:
For the sake of leaving you lots of PRs for after the long weekend. I think this will improve your quality of life for all the documentation work.

It will require you to enable PR previews https://docs.readthedocs.io/en/latest/pull-requests.html but besides that I think this should be setup. It will build the docs and add a link to ALL PRs. Optionally we could just have that happen for ones that change the docs (see commented block).

The basic config seemed like it satisfied our needs. Once we merge https://github.com/sandialabs/pyttb/pull/163 it will also allow for verifying the notebooks run and look nice when we change them.


dmdunla:
Enabled `Build pull requests for this project` on RTD.

@ntjohnson1 Can you create an account on RTD so I can add you as a maintainer. We do not have any backup maintainer ATM.

dmdunla:
@ntjohnson1 It looks like RTD build failed. Can you take a look at this?

ntjohnson1:
Yes to both. It looks like our RTD is configured to build with python3.7 but our package supports a minimum of 3.8. I should be able to look into where the configuration gets picked up from later this afternoon when I make an account.

dmdunla:
I think we need to create a `.readthedocs.yaml` file and add an option for `build.tools.python`:

https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python

PR #186, opened 2023-07-07
Ktensor lint type

ntjohnson1:
Add typing and linting coverage to ktensor re: https://github.com/sandialabs/pyttb/issues/63

### DIFF
--- a/pyttb/ktensor.py
+++ b/pyttb/ktensor.py
@@ -12,14 +12,10 @@ import numpy as np
 import scipy.sparse.linalg
 
 import pyttb as ttb
-from pyttb.pyttb_utils import (
-    isvector,
-    isrow,
-    tt_ind2sub
-)
+from pyttb.pyttb_utils import isrow, isvector, tt_ind2sub
 
 
-class ktensor():
+class ktensor:
     """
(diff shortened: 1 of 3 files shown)


pyttb/ktensor.py
```diff
@@ -49,25 +60,24 @@ class ktensor(object):
 
     __slots__ = ("weights", "factor_matrices")
 
-    def __init__(self, factor_matrices=None, weights=None, copy=True):
+    def __init__(
+        self,
+        factor_matrices: Optional[List[np.ndarray]] = None,
+        weights: Optional[np.ndarray] = None,
+        copy: bool = True,
```
dmdunla, line 67: Should all optional parameters use `Optional[...]`? Why is `copy` not typed as `Optional[bool]`?

ntjohnson1: `Optional[SomeType]` is equivalent to `Union[NoneType, SomeType]`. For `copy` we only process bools but provide a default value. Slightly different interpretations of optional.

dmdunla: Got it! That makes sense.

PR #190, opened 2023-07-07
Finish Core Typing

ntjohnson1:
This covers linting and typing for TTENSOR, TENMAT, and PYTTB_UTILS.

For users of the data structures as a toolbox this should be sufficient typing coverage to be interesting/useful. So additionally I added [py.typed](https://peps.python.org/pep-0561/) to the package so next release it will be incorporated in the wheel. With the ktensor changes next release probably needs to be 1.7 anyway.

I updated the ticket to cover the [algorithms](https://github.com/sandialabs/pyttb/issues/63) next. There are definitely quality of life improvements around centralizing some common logic and types to make consistency easier. But IMO the primary value is getting initial coverage everywhere so the lay of the land is a bit clearer.

----
:books: Documentation preview :books:: https://pyttb--190.org.readthedocs.build/en/190/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,6 +4,12 @@
 # Required
 version: 2
 
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.8"
+
 # Build documentation in the "docs/" directory with Sphinx
 sphinx:
    configuration: docs/source/conf.py
@@ -17,7 +23,6 @@ sphinx:
 # to build your documentation
 # See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
 python:
-  version: 3.8
   install:
     - method: pip
       path: .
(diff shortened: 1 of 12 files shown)


dmdunla (changes requested):
Looks like the doc build had an error. Now we see it due to the last PR merge. But I was unable to understand the error. Is this related to the notebook metadata or something else?

ntjohnson1:
Hmm I'll work on debugging it later this weekend. I can reduce the size of the problem and increase the myst timeout to see if there is just variability in the solve time causing this to be flaky.

ntjohnson1:
> Is this related to the notebook metadata or something else?

It looks like late Friday brain was making connections based on random changes. I added the description to the commits I added but:
* Reduced CP_APR size
* Increased myst timeout
* Added numpy random seed (to minimize build to build variability)
* Added cp_apr timeout argument

I cleared the myst/sphinx cache then re-ran 5 times locally and it passed every run with time to spare. So hopefully this resolves.

dmdunla (approved):

PR #199, opened 2023-07-13
197 algorithm profiling

DeepBlockDeepak:
Preliminary profiler notebook within `/profiling`, which displays top 10 slowest function calls as they occur in either `cp_als()` or `cp_apr()`, depending on which algorithm the user wants to test. The `ranks` list is looped over to test the functions with tensors of ranks 2,3 and finally 4. The data for the tests is read from `/profiling/data/*.tns`. 

### To test
The bottom code block allows you to choose which algorithm to test. At the moment, `cp_apr` takes ~5m30s while `cp_als` takes about 10s. 
```python
profile(test_files, ranks, 'cp_apr')
profile(test_files, ranks, "cp_als")
```


DeepBlockDeepak:
Thanks!
* I originally was testing on the `/tests/data/*.tns` files (without any input validation), but forgot to remove the ktensor check once you supplied the `/profiling/data/` directory.
* Will do on the rest. Thanks for the feedback!

dmdunla:
Looking good!

Here are some suggestions:
- Instead of commenting out all but one call in the last cell, put each call in a separate cell
- To get more runtime/computation out of each run, your can change `stoptol`, `maxiters`, `maxinneriters` (where available)
- Also, test the different `algorithm` choices in `cp_apr`: `mu` (default), `pqnr`, `pdnr`

DeepBlockDeepak:
Latest commit includes a new file `cp_and_tucker_algorithms_profiling_with_gprof2dot.ipynb` where the profiling stats are created in a generated sub-directory `/profiling/profiling_output_files`. 

### Example:

- When `profile(test_files, ranks, "hosvd", tol=1e-4)` is executed in the notebook, the data-input file `tensor_dense_integer_size_1000_3_3_rng_1.tns` will be used during computation at some point. The resulting profile stats will be created and saved as `./profiling_output_files/tensor_dense_integer_size_1000_3_3_rng_1_hosvd.pstats` relative to our notebook's location. Notice that the last part of the basename of that filepath includes a reference to the algorithm that was used. 
  -  Similarly, when `cp_als()` uses the same input tensor file, the resulting profile stats are saved like: `./profiling_output_files/tensor_dense_integer_size_1000_3_3_rng_1_cp_als.pstats`.

### To create corresponding .png files for the pstats data:
- Right now, the unsophisticated way I'm running this is to first navigate to the `/profiling` directory, then running the following command in Terminal:
```bash
mkdir -p output_images && for file in profiling_output_files/*.pstats; do gprof2dot -f pstats $file | dot -Tpng -o output_images/$(basename $file .pstats).png; done
```
This generates the `/profiling/output_images` directory, where the created .png profile visuals are stored. I believe I had to install both `graphviz` to my system and package install `gprof2dot` beforehand, a la [this introduction](https://nesi.github.io/perf-training/python-scatter/profiling-cprofile).

DeepBlockDeepak:
Note: typo in cell 20 should be: `profile(test_files, ranks, "tucker_als", stoptol=1e-5, maxiters=2000)` , with `tucker_als` being tested, instead of "cp_als".

DeepBlockDeepak:
The current visual profiling output is the result the pstats from testing with cell 11 :  `profile(test_files, ranks, "tucker_als")`.

dmdunla:
@DeepBlockDeepak Can you add the profiling output with images for all 4 algorithms?

DeepBlockDeepak:
@dmdunla The images exist inside of `/profiling/gprof2dot_images/<algo_name>` per each algorithm.

PR #200, opened 2023-07-14
Gcp fg

ntjohnson1:
Resolves https://github.com/sandialabs/pyttb/issues/198 and is the beginning of https://github.com/sandialabs/pyttb/issues/58
This adds `tensor.mttkrps` since that is used in `tt_gcp_fg`

Organization notes:
- I assume we are waiting to cut a new release until 2.0 but I expect some of this to shuffle a bit as I land other pieces so would prefer if we are going to land a 1.X.Y before then we note the gcp pieces are experimental
- Since GCP is a fairly large fairly standalone component I think a subdirectory or some other organization is helpful (at least for me)
- For testing the handles I basically just implemented them twice, once element by element which should be correct by inspection on the test side then vectorized with numpy on the source side. In theory we could calculate point golden values but that seems harder to maintain/inspect

Interface notes:
- I placed all the objective and gradient handles in their own file to validate independently of fg.
- fg_setup should be basically the same. Jeremy had some suggestions on easier workflow but I figured for now keeping things fairly aligned reduces confusion. Can iterate on internals in follow up.
- tt_gcp_fg - I renamed to evaluate inside of pyttb.gcp.fg since we are evaluating the function handles. By defaulting to none for the handles we can drop the flags to compute on or the other. I dropped the vectorization option. That might be a utility I add in later steps but for `tt_gcp_fg` I think it just makes the interface more confusing.

----
:books: Documentation preview :books:: https://pyttb--200.org.readthedocs.build/en/200/

Issue #58, opened 2023-02-23
implement gcp_opt


dmdunla:
- [x] Samplers [PR](https://github.com/sandialabs/pyttb/pull/201)
   - ~[Random Sampler](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_random_sample.m)~ (Looks like only for cparls)
   - [x] [Sample Non-zeros](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_sample_nonzeros.m)
   - [x] [Semistratified Sampling](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_sample_semistrat.m)
   - [x] [Stratified Sampling](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_sample_stratified.m)
   - [x] [Uniform Sampling](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_sample_uniform.m)
   - [x] [Sample Zeros](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_sample_zeros.m)
   - ~[F Sampler Setup](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_fsampler_setup.m)~ (Also looks like only cparls)
- [x] Objectives and gradients [PR](https://github.com/sandialabs/pyttb/pull/200) and secondary [ticket](https://github.com/sandialabs/pyttb/issues/198)
   - [x] [FG Setup](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_gcp_fg_setup.m)
   - [x] [Gradients and loss](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_gcp_fg.m)
   - [x] [Estimate Gradients and loss](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/tt_gcp_fg_est.m)
(https://docs.scipy.org/doc/scipy/reference/optimize.minimize-lbfgsb.html)
- [x] Driver
   - [x] [gcp_opt](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/gcp_opt.m)
- [x] Optimizers  
   - [x] [LBFGSB](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/gcp_opt.m#L459) Maybe [scipy]
   - [x] SGD
   - [x] Adam
   - [x] Adagrad

PR #201, opened 2023-07-14
Gcp samplers

ntjohnson1:
Sampler support for gcp_opt. Related to but doesn't complete https://github.com/sandialabs/pyttb/issues/58.

The tests are probably a little light but I don't have any clever ideas at the moment on how to better validate them. I cover the major portions but for example uniform could be sampling with a totally different distribution and still pass.

I updated my previous comment here, and dropped common commits so this is now distinct from https://github.com/sandialabs/pyttb/pull/200. Merge order shouldn't matter.

----
:books: Documentation preview :books:: https://pyttb--201.org.readthedocs.build/en/201/

### DIFF
--- a/pyttb/gcp/samplers.py
+++ b/pyttb/gcp/samplers.py
@@ -150,3 +150,37 @@ def uniform(
     vals = data[subs]
     wgts = (np.prod(data.shape) / samples) * np.ones((samples,))
     return subs, vals, wgts
+
+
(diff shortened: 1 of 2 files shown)


dmdunla (approved):
We can work on testing more when we put together the algorithm tutorials. THat may be a good user-based perspective to help drive useful tests.

PR #204, opened 2023-07-19
Gcp fg sptensor

jeremy-myers:
* Adds sptensor support to gcp/fg.py.
* Adds a test of all function and gradient handles with function values and gradient values computed with MATLAB. Also adds a test sptensor and tensor that correspond to these values.

----
:books: Documentation preview :books:: https://pyttb--204.org.readthedocs.build/en/204/


dmdunla:
@jeremy-myers can you check you recent commits? It looks like there are conflicts.

pyttb/gcp/fg.py
```diff
@@ -13,42 +13,42 @@
 @overload
 def evaluate(
     model: ttb.ktensor,
-    data: ttb.tensor,
-    weights: Optional[np.ndarray],
+    data: Union[ttb.tensor, ttb.sptensor],
     function_handle: Literal[None],
     gradient_handle: function_type,
+    weights: Optional[np.ndarray],
```
ntjohnson1, line 19: Is there a reason you re-arranged this interface? I made the ordering aligned between estimate and evaluate to keep them as similar as possible for ease of use. I didn't create a specific test for that though, so it's not strongly enforced at the moment.

tests/gcp/test_fg.py
```diff
@@ -1,14 +1,19 @@
...
+import scipy
 
 import pyttb as ttb
+from pyttb.gcp import fg_setup
 from pyttb.gcp.fg import evaluate
+from pyttb.gcp.handles import Objectives
 
 
 def test_evaluate():
-    # No function or gradient handle
```
ntjohnson1, line 11: This looks like a bad merge. I think I added the comment for more description.

tests/gcp/test_fg.py
```diff
@@ -1,14 +1,19 @@
 from __future__ import annotations
 
+from inspect import Parameter, signature
+from math import exp, log, pi
+
```
ntjohnson1, line 5: I think all these imports are also a bad merge. They don't appear to be used. I don't have pylint setup on our tests or an auto-unused import pruning check setup.

tests/gcp/test_fg.py
```diff
@@ -51,3 +56,72 @@ def no_op(data, model):
...
+        -61.434016498399011,
+        -125.9778934454338,
+        -26.214062308801779,
+        9751.193889205471,
+        -7705.253241233131,
+        -3844.118857614557,
+        1.659013125911521,
+        18.038214072232726,
+        438.0174726326384,
+    ]
```
ntjohnson1, line 78: Small preference would be if we had a way to independently generate these dynamically in a way independent of the source code that would be nice. For example, if MATLAB changes we won't know these went stale or how to generate them.

If that's not feasible at the moment then a smaller change that probably makes maintenance easier is making these maps from name to value. That way if the order of `Objectives` happens to need to change or we add a new objective MATLAB doesn't have it doesn't automatically break this test.

tests/gcp/test_fg.py
```diff
@@ -51,3 +56,72 @@ def no_op(data, model):
     )
     assert isinstance(f, float)
     assert all(isinstance(g_i, np.ndarray) for g_i in g)
+
+
+def test_evaluate_sptensor():
+    model = ttb.import_data("tests/data/ktensor_low_rank_small.tns")
+    data = ttb.import_data("tests/data/sptensor_low_rank_small.tns")
```
ntjohnson1, line 63: Also I'd prefer a pytest fixture that generates these stable values rather than checking in the files. But not a strong opinion on that one since we already have some other test data files checked in.

ntjohnson1 (commented):
A few notes about cleanup/future maintenance that would be nice to be addressed prior to approval, but otherwise looks good to me. Will leave final approval and merge to Danny.

dmdunla:
> A few notes about cleanup/future maintenance that would be nice to be addressed prior to approval, but otherwise looks good to me. Will leave final approval and merge to Danny.

@jeremy-myers I’ll let you take a first pass to resolve comments from @ntjohnson1 .

Issue #172, opened 2023-07-01
Implement `__getitem__` for `ttensor`


DeepBlockDeepak:
`__getitem__()` needs to be implemented for the ttensor class.

```python
>>> import pyttb as ttb
>>> import numpy as np
>>> np.random.seed(0)
>>> core = ttb.tensor.from_data(np.random.rand(3,2,1), shape=(3,2,1)) # The core tensor.
>>> U = [np.random.rand(5,3), np.random.rand(4,2), np.random.rand(3,1)] # The factor matrices.
>>> X = ttb.ttensor.from_data(core, U) # Create the ttensor.
>>> X[1] # No __getitem__()

TypeError: 'ttensor' object is not subscriptable
```

ntjohnson1:
Looking at the MATLAB documentation `X[index]` is simply a shorthand for `X.u[index]`. My vote would be to avoid the redundancy and require the explicit indexing into either `X.core` or `X.u`. Making this same decision for `ktensor` should also resolve https://github.com/sandialabs/pyttb/issues/165. In general if we can have one an only one way to do the right thing that's preferable.

CC: @dmdunla for thoughts

dmdunla:
I do not see `X(index)` working in MATLAB. It does provide `X{index}`, but we are not using cell arrays in `pyttb`. 

Thus, I agree with Nick that we should not support indices into a `ttensor`, just support indexing into the core and individual factor matrices.

PR #205, opened 2023-07-20
Updating ttensor tutorial after closing #172.

dmdunla:


dmdunla:
BTW, I cleared the output and several metadata fields that have been changing every time a new developed commits:
```bash
nbstripout docs/source/tutorial/class_ttensor.ipynb \
--extra-keys "metadata.language_info metadata.vscode metadata.kernelspec"
```

We could consider creating an issue about this when we are ready to support auto nbstripout (or something else)

ntjohnson1 (approved):
Approving since the core motivation is complete. However there are a few cleanup things that would be nice pending bandwidth here or as follow up.

- https://github.com/sandialabs/pyttb/blob/5517f34004f5ebe080b43856f0461109b02d54be/pyttb/ttensor.py#L151-L152
   - We probably want a `\n` after `Core is a` then indent the core string `\t\t` to match the factors.
- In [this](https://github.com/sandialabs/pyttb/blob/202b6fbae30d715bd37b7b8c74c6ad5fb40eadaa/docs/source/tutorial/class_ttensor.ipynb#L82) cell we probably want a trailing `Y` so we print the ttensor for sparse and dense cores
-  [This](https://github.com/sandialabs/pyttb/blob/202b6fbae30d715bd37b7b8c74c6ad5fb40eadaa/docs/source/tutorial/class_ttensor.ipynb#L139) says `cell array of matrices` where it is a list of matrices
- I thought I mentioned it but I don't think I saw it on the issue you closed for `__get_item__` do we want to rename `ttensor.u` to match `ktensor`? Either to `factor_matrices` or potentially making both just `factors` to be slightly less verbose?

As for nbstripout I do think with our active work on the notebooks it would be great to integrate, or at least get a ticket up. I doubt I'll have time to take a stab at it.

PR #206, opened 2023-07-21
Gcp opt

ntjohnson1:
Here is the initial land of the variety of gcp_opt components. The implementations should match pretty closely to MATLAB however in MATLAB there are so many parameters. I tried to group things in a few different ways. This is documented in the code/sphinx but wanted to elaborate here in case it is unclear so we can address earlier.
- `Sampler` class to be responsible for all of our different random draws
- `Optimizers` classes to have less branching inside of gcp_opt. IMO this should also make it easier to drop in other optimizers and keep things organized.

Major remaining tasks I can begin filing tickets for are:
- Nicer print statements/capturing all our parameters per class
- More examples/elaboration in our docstrings.
- Quick profile for hot spots since just setting up the small tests things seemed to run slower than I would expect (but I didn't look into that at all)
- Thoughts around more involved numeric validation (probably related to the above profiling where we run a few things to convergence)

----
:books: Documentation preview :books:: https://pyttb--206.org.readthedocs.build/en/206/

### DIFF
--- a/docs/source/algorithms.rst
+++ b/docs/source/algorithms.rst
@@ -5,5 +5,6 @@ Algorithms
 
     cpals.rst
     cpapr.rst
+    gcpopt.rst
     hosvd.rst
     tuckerals.rst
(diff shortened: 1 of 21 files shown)


ntjohnson1:
FYI that the preliminary gcp opt landed: @jeremy-myers @tgkolda

PR #211, opened 2023-08-13
Ruff addition

ntjohnson1:
This adds ruff to show 1) how fast it runs (running it in addition to everything else shouldn't really add additional runtime) and 2) some of the subtle differences `noqa` with codes to skip individual blocks that don't conform. I don't think people were particularly concerned about the specific rules we were enforcing before but wanted to propose the addition of ruff before I went through the larger effort of:
- Removing pylint/isort
- Updating the dev docs
- Updating our pre-commit integration

There are a variety of additional ruff [rule](https://beta.ruff.rs/docs/rules/) sets if we decided there were additional things we'd like the linter to capture (such as sanity checking our [doc strings](https://beta.ruff.rs/docs/faq/#does-ruff-support-numpy-or-google-style-docstrings) in addition to doc test) but I think the current rules cover the most important pieces.

----
:books: Documentation preview :books:: https://pyttb--211.org.readthedocs.build/en/211/


dmdunla (approved):
In general, I like this move. The addition of the caveats per function (e.g., `# noqa: PLR0913`) is distracting in the code, IMHO. It looks like there is a way to set some package params to address some if not all of these caveats without cluttering the source. For users who just call the algorithms, this is not a big deal. For users who want to use `pyttb` as a research framework, this adds unnecessary cognitive load.

ntjohnson1:
> In general, I like this move. The addition of the caveats per function (e.g., `# noqa: PLR0913`) is distracting in the code, IMHO. It looks like there is a way to set some package params to address some if not all of these caveats without cluttering the source. For users who just call the algorithms, this is not a big deal. For users who want to use `pyttb` as a research framework, this adds unnecessary cognitive load.

My general preference is to provide overly strict rules with the option to opt-out of them selectively. Generally, it is much easier to have the incremental checks during development rather than a big refactor later. Most of the linting rules are to catch subtle bugs, point out code smells, and make suggestions for more manageable code. The most common function by function exemptions were PLR0912, PLR0913, and PLR0915 (too-many-: branches, args, statements) which are configurable by (with listed defaults)
```bash
[tool.ruff.pylint]
max-args = 5
max-branches = 12
max-statements = 50
```
I am open to any values you think are more reasonable, if you feel strongly we should disable the rules globally, or ruff also has the option to disable on a per file basis. Resolving the places we opt out of the rules will _probably_ help in the long term. (ex: CP_APR currently has 18 different arguments which probably complicates its use). So, as with everything, it is a balance. A variety of different options available, and we can always punt for now and see if contributors run into issues. (ruff has a handy `ruff check <path> --add-noqa` if that reduces the mental load aspect for reasonable opt-outs)

Issue #209, opened 2023-08-10
ttensor: align name of factor matrices with ktensor


dmdunla:
Currently, `ttensor.u` stores the factor matrices of Tucker tensors. To align with `ktensor` factor matrices, we should change the name to `ttensor.factor_matrices`.

dmdunla:
Closed by #213

Issue #188, opened 2023-07-07
Sptensor: add support for constructing from data without copying


dmdunla:
Follow up to 
- #145 

Follow changes from #182 for `ktensor` as a guide.

Constructors:
- Update: `__init__`: (replaces `from_data`, adds copy parameter)
- Update: `from_function`
- Update: `from_aggregator`
- Remove: `from_tensor_type`
- Remove: `from_data`

Add methods:
- `copy`

dmdunla:
Closed with #213

Issue #187, opened 2023-07-07
Tensor: add support for constructing from data without copying


dmdunla:
Follow up to 
- #145 

Follow changes from #182 for `ktensor` as a guide.

Constructors:
- Update: `__init__`: (replaces `from_data`, add copy parameter)
- Update: `from_function`
- Remove: `from_tensor_type`
- Remove: `from_data`

Add methods:
- `copy`

dmdunla:
Closed with #213

Issue #145, opened 2023-06-09
Add support for constructing tensor/ktensor from data without copying


etphipp:
This is an issue to discuss adding support for creating `tensor` and `ktensor` objects (and probably others) without copying the original data the object is derived from.  This will improve efficiency, especially for third-party codes that are trying to integrate with pyttb (such as GenTen).  I have a PR that could be submitted that implements one proposed solution by adding a `copy` parameter to the `from_data` methods of `tensor` and `ktensor` (defaulting to `True` to preserve existing behavior).  Note also that the various functions for creating various types of tensors from data are not entirely consistent on whether the data is copied or not.

Another reasonable solution in mind, and one that I would argue is preferable, is to never explicitly copy the data in these functions, and if the user wants a copy, to call the `copy` method of the class instead.

dmdunla:
I like the idea of having all `from_data` methods not making copies and providing an explicit copy method. This is essentially what `ktensor.from_tensor_type`.

etphipp:
I can start working on a PR for that.  Scanning through the code, it looks like the arguments to `from_data` are often copied before passing to `from_data` anyway.  And `ktensor` already has a `copy` method.

dmdunla:
Here are the constructors with suggested behavior:

- ` tensor`
  - `from_data`: no copy
  - `from_tensor_type`: copy (to support conversion)
  - ``from_function`: new
- `ktensor`
  - `from_data`: no copy
  - `from_tensor_type`: copy (for consistency)
  - `from_factor_matrices`:  no copy (creates weights of 1 and then calls `from_data)
  - `from_function`: new
  - `from_vector`: copy (data is reshaped)
- `sptensor`
  - `from_data`: no copy
  - `from_tensor_type`: copy (for consistency))
  - ``from_function`: new
  - `from_aggregator`: new/copy (as data may be aggregated and not be the same size afterwards)
- `ttensor`
  - `from_data`: no copy
  - `from_tensor_type`: copy (for consistency)
- `tenmat`
  - `from_data`: copy
  - `from_tensor_type`: copy (for consistency)

dmdunla:
Because `from_tensor_type` has different behaviors, the consistency will be challenging. 

Suggestion: remove `from_tensor_type` constructors and create either `to_*` or `as_*` methods that map from one class to another. This would limit the behavior to conversion only, as opposed to the currnt behaviors of both conversion and calling from_data, sometimes copying other times not.

[...]

PR #213, opened 2023-08-19
Constructor update

ntjohnson1:
Addresses #187 #188 (and ttensor but I didn't bother with a ticket for that) and #209. Filed #214 as follow up because tenmat's relationships between the constructors is kind of weird, so needs more thought. I followed the design discussed in #145 which I believe was the intention of the first two issues.

There were also three commits https://github.com/sandialabs/pyttb/pull/213/commits/ca9629bf41462b0ebbed56cc059c0e9d8a2cc9d1, https://github.com/sandialabs/pyttb/pull/213/commits/862201b070355b1b7d4dbf8ac6a2465c4e97e795, https://github.com/sandialabs/pyttb/pull/213/commits/2655c829db75450dcff91bc4e0a2a315d295b754 related to simplifying some edge cases in ruff. Basically, we replaced isort with ruff but weren't applying the import sorting/removal of redundant imports to our tests. Also ruff can detect and fix unused variables etc which could be useful if we realize our tests aren't actually doing what we expect.

A few notes for where I set copy=False already. I didn't look at our profiling at all to see if this helped in our performance. I didn't do an exhaustive analysis for where we can skip the copies. Generally I looked for obvious usages where copies were redundant:
* All inputs are newly allocated values
* The re-used inputs are imutable (such as shape) so we don't have to worry about the reference issue

~While the content is mostly unrelated I expect this to clash with #212 but they both are based off latest master so feel free to merge whichever you prefer first and I'll resolve the conflicts.~(Resolved)

----
:books: Documentation preview :books:: https://pyttb--213.org.readthedocs.build/en/213/


dmdunla:
Just updated #212. Please update as needed.

ntjohnson1:
Great. I rebased and did a quick check so this should be ready to go now. I elaborated a bit more on the PR description in case that is useful, but feel free to let me know if anything seems weird or I misinterpreted the direction we were discussing for the interfaces.

dmdunla:
I notice differences from `ktensor` but only in the docs ...

Can you align the `docs/source/*.rst` files for the tensor classes with `ktensor.rst`. The docs show `pyttb.ktensor()` as the class constructor, whereas `pyttb.sptensor.sptensor()` is the class constructor with a method called `__init__`. The same in the latter case for `tensor` and `ttensor`.  With the changes the documentation makes them appear very different, and I confused these differences for implementation differences. I think the main difference is the exclusion of `__init__`, but this should be confirmed.

pyttb/tensor.py
```diff
@@ -36,19 +36,12 @@ class tensor:
     data: np.ndarray
```
dmdunla, line 36: I do not think we want these class variables. now that we have move our constructor into `__init__`. I think that `self.data` and `self.shape` defined in `__init__` makes them instance variables, which is what we want. I don't think I fully understand this, but it seems that if we define instance variables in `__init__` that have the same name as the class variables (as defined here), we never access the class variable -- thus it is not needed.

ntjohnson1: Ya that's a good catch. I don't think these should have ever been class variables and I was just using them when I was fighting adding typing.

ntjohnson1:
> I notice differences from `ktensor` but only in the docs ...
> 
> Can you align the `docs/source/*.rst` files for the tensor classes with `ktensor.rst`. The docs show `pyttb.ktensor()` as the class constructor, whereas `pyttb.sptensor.sptensor()` is the class constructor with a method called `__init__`. The same in the latter case for `tensor` and `ttensor`. With the changes the documentation makes them appear very different, and I confused these differences for implementation differences. I think the main difference is the exclusion of `__init__`, but this should be confirmed.

I added two commits for this since there are some stylistic preferences (details are in the commits but adding more info in case it is clearer). For the first I dropped `undoc-members` since if we haven't documented things I think we should skip them (this avoids needing to explicitly skip `factor_matrices` and other class members). Then I un-excluded `__init__` so we actually got the documentation for that.

In the second commit I merged the `__init__` docstring with the class doc string since that  was more similar to what ktensor looked like before and I think is the desired style.

Since this brings all our tensor types into the same format in the future copy pasting should make new types match as well.

PR #216, opened 2023-08-25
20 tensor complete documentation for all methods

dmdunla:
Updating docstrings, doctests.

----
:books: Documentation preview :books:: https://pyttb--216.org.readthedocs.build/en/216/

### DIFF
--- a/docs/source/sptensor.rst
+++ b/docs/source/sptensor.rst
@@ -1,8 +1,8 @@
 pyttb.sptensor
 ----------------------
 
-.. autoclass:: pyttb.sptensor
+.. automodule:: pyttb.sptensor
     :members:
     :special-members:
-    :exclude-members: __dict__, __weakref__, __slots__, __init__
+    :exclude-members: __dict__, __weakref__, __slots__
     :show-inheritance:
--- a/docs/source/sptensor.rst
+++ b/docs/source/sptensor.rst
@@ -1,8 +1,8 @@
 pyttb.sptensor
 ----------------------
 
-.. automodule:: pyttb.sptensor
+.. autoclass:: pyttb.sptensor
     :members:
     :special-members:
-    :exclude-members: __dict__,__weakref__, __slots__, __init__
+    :exclude-members: __dict__, __weakref__, __slots__, __init__
     :show-inheritance:
--- a/docs/source/sptensor.rst
+++ b/docs/source/sptensor.rst
@@ -4,5 +4,5 @@ pyttb.sptensor
 .. automodule:: pyttb.sptensor
     :members:
     :special-members:
-    :exclude-members: __dict__, __weakref__, __slots__
+    :exclude-members: __dict__, __weakref__, __slots__, __init__
     :show-inheritance:
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -59,7 +59,7 @@ class tensor:
         copy: bool = True,
     ):
         """
-        Creates a :class:`pyttb.ktensor` from a :class:`numpy.ndarray`
+        Creates a :class:`pyttb.tensor` from a :class:`numpy.ndarray`
         Note that 1D tensors (i.e., when len(shape)==1) contains a data
         array that follow the Numpy convention of being a row vector.
 
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -1,4 +1,4 @@
-"""Dense Tensor Implementation"""
+"""Classes and functions for working with dense tensors."""
 # Copyright 2022 National Technology & Engineering Solutions of Sandia,
 # LLC (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the
 # U.S. Government retains certain rights in this software.
@@ -42,27 +42,36 @@ class tensor:
         copy: bool = True,
     ):
         """
-        Creates a tensor from explicit description. Note that 1D tensors (i.e.,
-        when len(shape)==1) contains a data array that follow the Numpy convention
-        of being a row vector, which is different than in the Matlab Tensor Toolbox.
+        Creates a :class:`pyttb.ktensor` from a :class:`numpy.ndarray` 
+        Note that 1D tensors (i.e., when len(shape)==1) contains a data 
+        array that follow the Numpy convention of being a row vector.
 
         Parameters
         ----------
         data:
-            Tensor source data
+            Tensor source data.
         shape:
-            Shape of resulting tensor if not the same as data shape
+            Shape of resulting tensor if not the same as data shape.
         copy:
-            Whether to make a copy of provided data or just reference it
+            Whether to make a copy of provided data or just reference it.
 
-        Returns
+        Examples
         -------
-        Constructed tensor
+        Create an empty :class:`pyttb.tensor`:
 
-        Example
-        -------
-        >>> X = ttb.tensor(np.ones((2,2)))
-        >>> Y = ttb.tensor(np.ones((2,2)), shape=(4,1))
+        >>> T = ttb.tensor()
+        >>> print(T)
+        empty tensor of shape ()
+        data = []
+
+        Create a :class:`pyttb.tensor` from a :class:`numpy.ndarray`:
+        >>> T = ttb.tensor(np.array([[1,2],[3,4]]))
+        >>> print(T)
+        tensor of shape (2, 2)
+        data[:, :] = 
+        [[1 2]
+         [3 4]]
+        <BLANKLINE>
         """
         if data is None:
             # EMPTY / DEFAULT CONSTRUCTOR
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -59,36 +59,36 @@ class tensor:
         copy: bool = True,
     ):
         """
-        Creates a :class:`pyttb.tensor` from a :class:`numpy.ndarray`
-        Note that 1D tensors (i.e., when len(shape)==1) contains a data
-        array that follow the Numpy convention of being a row vector.
+                Creates a :class:`pyttb.tensor` from a :class:`numpy.ndarray`
+                Note that 1D tensors (i.e., when len(shape)==1) contains a data
+                array that follow the Numpy convention of being a row vector.
 
-        Parameters
-        ----------
-        data:
-            Tensor source data.
-        shape:
-            Shape of resulting tensor if not the same as data shape.
-        copy:
-            Whether to make a copy of provided data or just reference it.
+                Parameters
+                ----------
+                data:
+                    Tensor source data.
+                shape:
+                    Shape of resulting tensor if not the same as data shape.
+                copy:
+                    Whether to make a copy of provided data or just reference it.
 
-        Examples
-        -------
-        Create an empty :class:`pyttb.tensor`:
+                Examples
+                -------
+                Create an empty :class:`pyttb.tensor`:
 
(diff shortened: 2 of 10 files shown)


ntjohnson1:
Is this waiting on review from me or still WIP? I assume the sptensor bug came up when looking over our documentation stuff

dmdunla:
Thanks, still working ... I'll explicitly request a review when ready.

pyttb/tensor.py
```diff
@@ -42,33 +59,42 @@ def __init__(
         copy: bool = True,
     ):
         """
-        Creates a tensor from explicit description. Note that 1D tensors (i.e.,
-        when len(shape)==1) contains a data array that follow the Numpy convention
-        of being a row vector, which is different than in the Matlab Tensor Toolbox.
+        Creates a :class:`pyttb.ktensor` from a :class:`numpy.ndarray`
```
ntjohnson1, line 62:
```suggestion
        Creates a :class:`pyttb.tensor` from a :class:`numpy.ndarray`
```

pyttb/tensor.py
```diff
@@ -111,25 +137,42 @@ def from_function(
...
         """
-        Creates a tensor from a function handle and size
+        Construct a :class:`pyttb.tensor` whose data entries are set using
+        a function.
 
         Parameters
         ----------
-        function_handle:
-            Function to generate data to construct tensor
+        fun: function, required
```
ntjohnson1, line 145: The argument in tensor is `function_handle` same with `sptensor`. Unless you find this terribly verbose we should update ktensor to match. Also `required` is unnecessary in the description since there is no default provided.

pyttb/tensor.py
```diff
@@ -111,25 +137,42 @@ def from_function(
...
+            A function that can accept a shape (i.e., :class:`tuple` of
+            dimension sizes) and return a :class:`numpy.ndarray` of that shape.
+            Example functions include `numpy.random.random_sample`,
+            `numpy,zeros`, `numpy.ones`.
```
ntjohnson1, line 149:
```suggestion
            `numpy.zeros`, `numpy.ones`.
```

Same typo exists in ktensor.

pyttb/tensor.py
```diff
@@ -282,58 +352,64 @@ def contract(self, i: int, j: int) -> Union[np.ndarray, tensor]:
 
     def double(self) -> np.ndarray:
         """
-        Convert tensor to an array of doubles
+        Convert tensor to an :array of doubles.
```
ntjohnson1, line 355:
```suggestion
        Convert `:class:pyttb.tensor` to an `:class:numpy.ndarray` of doubles.
```

pyttb/tensor.py
```diff
@@ -220,39 +269,60 @@ def collapse(
         ## Form and return the final result
         return ttb.tensor(B, newshape, copy=False)
 
-    def contract(self, i: int, j: int) -> Union[np.ndarray, tensor]:
+    def contract(self, i1: int, i2: int) -> Union[np.ndarray, tensor]:
```
ntjohnson1, line 272: Is there a particular reason to prefer `i1, i2` over `i,j`? I don't have a strong rationale but `i`, `j` looks nicer to me.

pyttb/tensor.py
```diff
@@ -342,15 +418,30 @@ def find(self) -> Tuple[np.ndarray, np.ndarray]:
 
     def to_sptensor(self) -> ttb.sptensor:
         """
-        Contruct an :class:`pyttb.sptensor` from `pyttb.tensor`
+        Contruct a :class:`pyttb.sptensor` from `pyttb.tensor`
```
ntjohnson1, line 421:
```suggestion
        Contruct a :class:`pyttb.sptensor` from `:class:pyttb.tensor`
```

pyttb/tensor.py
```diff
@@ -361,20 +452,26 @@ def full(self) -> tensor:
         """
         return ttb.tensor(self.data)
 
-    def innerprod(self, other: Union[tensor, ttb.sptensor, ttb.ktensor]) -> float:
+    def innerprod(
+        self, other: Union[tensor, ttb.sptensor, ttb.ktensor, ttb.ttensor]
+    ) -> float:
         """
-        Efficient inner product with a tensor
+        Efficient inner product between a tensor and other `pyttb` tensor
```
ntjohnson1, line 459:
```suggestion
        Efficient inner product between a tensor and other `pyttb` tensors
```

pyttb/tensor.py
```diff
@@ -1889,60 +2359,89 @@ def __repr__(self):
...
+        Shape of resulting tensor.
 
     Returns
     -------
-    Constructed tensor
+    Constructed tensor.
 
-    Example
+    Examples
     -------
```
ntjohnson1, line 2374:
```suggestion
    --------
```

pyttb/tensor.py
```diff
@@ -42,33 +59,42 @@ def __init__(
...
-        -------
-        >>> X = ttb.tensor(np.ones((2,2)))
-        >>> Y = ttb.tensor(np.ones((2,2)), shape=(4,1))
+        >>> T = ttb.tensor()
+        >>> print(T)
+        empty tensor of shape ()
+        data = []
+
+        Create a :class:`pyttb.tensor` from a :class:`numpy.ndarray`:
+        >>> T = ttb.tensor(np.array([[1,2],[3,4]]))
```
ntjohnson1, line 85: You need a newline between the description and code for it to render correctly https://pyttb--216.org.readthedocs.build/en/216/tensor.html#pyttb.tensor

pyttb/tensor.py
```diff
@@ -282,58 +352,64 @@ def contract(self, i: int, j: int) -> Union[np.ndarray, tensor]:
...
     def exp(self) -> tensor:
         """
-        Exponential of the elements of tensor
+        Exponential of the elements of tensor.
 
         Returns
         -------
-        Copy of tensor data element-wise raised to exponential
+        Copy of tensor data wtih the exponential function applied to data
+        element-wise.
```
ntjohnson1, line 377: I think for the multiline returns it needs to be indendented to render correctly https://pyttb--216.org.readthedocs.build/en/216/tensor.html#pyttb.tensor.exp

pyttb/tensor.py
```diff
@@ -282,58 +352,64 @@ def contract(self, i: int, j: int) -> Union[np.ndarray, tensor]:
...
         """
-        FIND Find subscripts of nonzero elements in a tensor.
+        Find subscripts of nonzero elements in a tensor.
 
-        S, V = FIND(X) returns the subscripts of the nonzero values in X and a column
-        vector of the values.
+        Returns
+        -------
+        Array of subscripts of the nonzero values in the tensor and a column
+        vector of the corresponding values.
```
ntjohnson1, line 395: Same indentations rendering note

pyttb/tensor.py
```diff
@@ -1796,6 +2238,16 @@ def __rtruediv__(self, other):
         Returns
         -------
         :class:`pyttb.tensor`
+
+        Examples
+        --------
+        >>> T = ttb.tensor(np.array([[1, 2], [3, 4]]))
+        >>> np.set_printoptions(precision=8)
```
ntjohnson1, line 2245: An alternative is to use the `...` doctest description to avoid comparing the full float. I think we use that a variety of other places, for example nvecs

docs/source/sptensor.rst
```diff
@@ -1,8 +1,8 @@
 pyttb.sptensor
 ----------------------
 
-.. automodule:: pyttb.sptensor
+.. autoclass:: pyttb.sptensor
```
ntjohnson1, line 4: Switching automodule to autoclass drops the auto documentation for the `tt_to_sparse_matrix` and `tt_from_sparse_matrix`. If these are going away/mostly internal then that's ok. Mostly just an fyi

pyttb/tensor.py
ntjohnson1: Hmm I see later on we used `fun` instead of the verbose `function_handle` in tensor. We should probably unify globally or file an issue to do so.

ntjohnson1 (commented):
Looks good to me. A variety of small fixes/notes.

One global style note. I think it is nice to have full print outs of tensor contents to be explicit, however it does add to maintenance efforts if we ever change the format of the display. Luckily doctest will catch it pretty immediately. I guess since you are the one who just did the doc updates you are in the best position to state if that's a worthwhile tradeoff.

dmdunla:
> Looks good to me. A variety of small fixes/notes.
> 
> One global style note. I think it is nice to have full print outs of tensor contents to be explicit, however it does add to maintenance efforts if we ever change the format of the display. Luckily doctest will catch it pretty immediately. I guess since you are the one who just did the doc updates you are in the best position to state if that's a worthwhile tradeoff.

I went back and forth on this. I think this provides users with examples of what they should see. Since we clear the tutorial notebooks, these docs provide a way to see what pyttb does before you start running it (or while you are trying to determine if it is doing what you want).

docs/source/sptensor.rst
dmdunla: These will go away once sptenmat is in place. I tried to align with ktensor, which provided cleaner docs, IMHO.

pyttb/tensor.py
dmdunla: That did not work in all places, but I can try again. I am using Github codespaces (while traveling) and there seem to be some differences from my typically VS code setup.

pyttb/tensor.py
dmdunla: I was trying to be explicit with the ordering. This follows papers that that move away from i,j,k that is used for 3-way tensors.

docs/source/sptensor.rst
dmdunla: I went back to automodule to include the methods in each of the class files. The class constructor is a bit confusing now, though, at least to me. It states `pyttb.tensor.tensor(...)`, instead of `pyttb.tensor(...)` as it does when stating autoclass.

@ntjohnson1 any thoughts on this?

pyttb/tensor.py
ntjohnson1: This wasn't actually resolved but I just pushed a commit that should have fixed the formatting.

docs/source/sptensor.rst
ntjohnson1: No strong thoughts. If you think the autoclass is cleaner we can probably stick to that for now. This relates to discussions we've had in the past about slightly ambiguous code layout because of our flat structure mapping to the MATLAB version. Probably doesn't need to be resolved for this PR though. We could also pull out some of the other methods from the class files and think about a better place for them so we could use autoclass AND capture the documentation for these methods.

Some more code organization thoughts:
* It's a little confusing we use tensor generically for all multi-dim data but also tensor specifically for dense tensors
* It adds to the ambiguity that we use lowercase class names (to match MATLAB)
* We add convenience imports to flatten the structure, but the docs probably should refer to the actual source location (I am not sure if there is an easy way to convince sphinx otherwise)

Related to the first two, in my repo where I started to play with a rust port (that I haven't touched in months). I was playing with a structure that looks like
```
ttb/
   tensors/
      Dense
      Kruskal
      Sparse
      Tucker
```

ntjohnson1 (approved):
I think for the intention of adding full docs coverage to tensor this PR looks good. There are a few open points, but those can probably get moved to a follow on ticket/discussion.

PR #219, opened 2023-08-26
21 sptensor complete documentation for all methods

dmdunla:
Updating sptensor documentation.

----
:books: Documentation preview :books:: https://pyttb--219.org.readthedocs.build/en/219/


dmdunla:
@ntjohnson1 Typing has me stuck on `sptensor.squeeze`. The output should allow `float` (like in `tensor.squeeze`), but when I changed the code, I got lost in the cascade of type mismatches. If you see something the latest branch commit associated with this error, please let me know. Thanks!

ntjohnson1:
I had an older version of numpy and it actually passed with no changes. However the squeeze with empty tuple indexing is definitely weird. I'm not sure how I ended up on that solution. I think what sptensor had originally is clearer (if no dim is greater than 1 we have a singleton so grab it directly). However, as you can see from my first commit this doesn't seem to work on the current version of numpy. Just calling item on a singleton array is also pretty clear and is backwards compatible, so that seems like a reasonable approach.

dmdunla:
@ntjohnson1 Thanks! I was thrown by the `np.array` output on `tensor.ttv`. This is a good fix. It avoids returns like this, `[[[[1]]]]`, which were coming out of `sptensor.ttv`, and it only changes the data type if we end up with a singleton tensor and just return a scalar.

Almost done with these docstrings.

pyttb/sptensor.py
```diff
@@ -440,43 +550,60 @@ def collapse(
...
         # Create Result
         if self.subs.size > 0:
             return ttb.sptensor.from_aggregator(
-                self.subs[:, remdims], self.vals, tuple(newsize), fun
+                self.subs[:, remdims], self.vals, tuple(newsize), function_handle
             )
         return ttb.sptensor(np.array([]), np.array([]), tuple(newsize), copy=False)
 
-    def contract(self, i: int, j: int) -> Union[np.ndarray, sptensor, ttb.tensor]:
+    def contract(self, i1: int, i2: int) -> Union[np.ndarray, sptensor, ttb.tensor]:
```
ntjohnson1, line 564: I think snake case feels a bit more pythonic. For our example S1 etc doesn't matter too much but for an interface it just looks a little awkward to me. Also zero indexing?
```suggestion
    def contract(self, i_0: int, i_1: int) -> Union[np.ndarray, sptensor, ttb.tensor]:
```

pyttb/sptensor.py
```diff
@@ -440,43 +550,60 @@ def collapse(
...
         """
         Contract tensor along two dimensions (array trace).
 
         Parameters
         ----------
-        i:
-            First dimension
-        j:
-            Second dimension
+        i1:
```
ntjohnson1, line 570:
```suggestion
        i_0:
```

pyttb/sptensor.py
```diff
@@ -440,43 +550,60 @@ def collapse(
...
+            First dimension.
+        i2:
```
ntjohnson1, line 572:
```suggestion
        i_1:
```

pyttb/sptensor.py
```diff
@@ -440,43 +550,60 @@ def collapse(
...
+        >>> subs = np.array([[1, 1, 1],[2, 2, 2]])
+        >>> vals = np.array([[0.5],[1.5]])
+        >>> shape=(4, 4, 4)
+        >>> S = ttb.sptensor(subs, vals, shape)
+        >>> S.contract(1,2)
+        sparse tensor of shape (4,) with 2 nonzeros
+        [1] = 0.5
+        [2] = 1.5
+        """
+        if self.shape[i1] != self.shape[i2]:
```
ntjohnson1, line 603:
```suggestion
        if self.shape[i_0] != self.shape[i_1]:
```

pyttb/sptensor.py
```diff
@@ -440,43 +550,60 @@ def collapse(
...
             assert False, "Must contract along equally sized dimensions"
 
-        if i == j:
+        if i1 == i2:
```
ntjohnson1, line 606:
```suggestion
        if i_0 == i_1:
```

pyttb/sptensor.py
```diff
@@ -485,13 +612,13 @@ def contract(self, i: int, j: int) -> Union[np.ndarray, sptensor, ttb.tensor]:
             return sum(self.vals[tfidx].transpose()[0])
 
         # Remaining dimensions after contract
-        remdims = np.setdiff1d(np.arange(0, self.ndims), np.array([i, j])).astype(int)
+        remdims = np.setdiff1d(np.arange(0, self.ndims), np.array([i1, i2])).astype(int)
```
ntjohnson1, line 615:
```suggestion
        remdims = np.setdiff1d(np.arange(0, self.ndims), np.array([i_0, i_1])).astype(int)
```

pyttb/sptensor.py
```diff
@@ -485,13 +612,13 @@ def contract(self, i: int, j: int) -> Union[np.ndarray, sptensor, ttb.tensor]:
...
 
         # Size for return
         newsize = tuple(np.array(self.shape)[remdims])
 
         # Find index of values on diagonal
-        indx = np.where(self.subs[:, i] == self.subs[:, j])[0]
+        indx = np.where(self.subs[:, i1] == self.subs[:, i2])[0]
```
ntjohnson1, line 621:
```suggestion
        indx = np.where(self.subs[:, i_0] == self.subs[:, i_1])[0]
```

pyttb/sptensor.py
```diff
@@ -520,26 +664,29 @@ def double(self) -> np.ndarray:
 
     def elemfun(self, function_handle: Callable[[np.ndarray], np.ndarray]) -> sptensor:
         """
-        Manipulate the non-zero elements of a sparse tensor
+        Apply a function to the non-zero elements of a sparse tensor.
```
ntjohnson1, line 667: If really motivated we have both non-zero and nonzero throughout the class. I think nonzero is more common.

pyttb/sptensor.py
```diff
@@ -671,12 +883,36 @@ def innerprod(
 
     def isequal(self, other: Union[sptensor, ttb.tensor]) -> bool:
         """
-        Exact equality for sptensors
+        Exact equality for sparse tensors, where all elements are exactly the
+        same in both tensors. Can be compared to another sparse tensor or
+        dense tensor.
```
ntjohnson1, line 888: Same note on redundant types.
```suggestion
        same in both tensors.
```

pyttb/sptensor.py
```diff
@@ -688,16 +924,43 @@ def isequal(self, other: Union[sptensor, ttb.tensor]) -> bool:
 
     def logical_and(self, B: Union[float, sptensor, ttb.tensor]) -> sptensor:
         """
-        Logical and with self and another object
+        Logical AND with a scalar or another sparse or dense tensor.
```
ntjohnson1, line 927: Numpys logical and is a function on two objects but their documentation takes a similar approach of not being too descriptive in the raw text, but instead the parameter annotations are more verbose.
```suggestion
        Logical AND with another object.
```

pyttb/sptensor.py
```diff
@@ -688,16 +924,43 @@ def isequal(self, other: Union[sptensor, ttb.tensor]) -> bool:
...
 
         Parameters
         ----------
-        B:
-            Other value to compare with
+        Scalar (float) value or other tensor to compare against.
```
ntjohnson1, line 931: By not indenting or providing the parameter name sphinx takes this whole line to be the parameter name.
If we wanted to we could rename `B` to `other` to match `isequal` which might be clearer.
```suggestion
        B:
           Other value to compare against.
```

pyttb/sptensor.py
```diff
@@ -671,12 +883,36 @@ def innerprod(
...
-        Exact equality for sptensors
+        Exact equality for sparse tensors, where all elements are exactly the
+        same in both tensors. Can be compared to another sparse tensor or
+        dense tensor.
 
         Parameters
         ----------
-        other:
-            Other tensor to compare against
+        Other tensor to compare against.
```
ntjohnson1, line 892: The rendering in sphinx looks weird when just a description rather than the parameter name is provided. For consistency I think we want to keep the param name even if there is just one.
```suggestion
        other:
           Other tensor to compare against.
```

pyttb/sptensor.py
```diff
@@ -688,16 +924,43 @@ def isequal(self, other: Union[sptensor, ttb.tensor]) -> bool:
...
         Parameters
         ----------
-        B:
-            Other value to compare with
+        Scalar (float) value or other tensor to compare against.
 
         Returns
         ----------
-        Indicator tensor
+        Indicator sparse tensor (values if 1.0 for True).
```
ntjohnson1, line 935:
```suggestion
        Indicator sparse tensor (values of 1.0 for True).
```

pyttb/sptensor.py
```diff
@@ -758,11 +1036,45 @@ def logical_or(
         self, B: Union[float, ttb.tensor, sptensor]
     ) -> Union[ttb.tensor, sptensor]:
         """
-        Logical OR for sptensor and another value
+        Logical OR for sptensor and a scalar or another sparse or dense tensor.
```
ntjohnson1, line 1039: Similar note as logical_and

pyttb/sptensor.py
```diff
@@ -796,16 +1108,44 @@ def logical_xor(
         self, other: Union[float, ttb.tensor, sptensor]
     ) -> Union[ttb.tensor, sptensor]:
         """
-        Logical XOR for sptensors
+        Logical XOR with a scalar or another sparse or dense tensor.
```
ntjohnson1, line 1111: Same note as logical_and

pyttb/sptensor.py
```diff
@@ -828,16 +1168,48 @@ def logical_xor(
...
+        :class:`pyttb.sptensor` containing only non-zero values that are
+        ones (1). The values in the :class:`pyttb.sptensor` corresponding
+        to the indices of the ones (1) in `W` will be returned as a column
+        vector.
 
         Parameters
         ----------
         W:
-            Mask tensor
+            Mask tensor to apply to ktensor.
```
ntjohnson1, line 1180:
```suggestion
            Mask tensor to apply to sparse tensor.
```

pyttb/sptensor.py
```diff
@@ -828,16 +1168,48 @@ def logical_xor(
 
     def mask(self, W: sptensor) -> np.ndarray:
         """
-        Extract values as specified by a mask tensor
+        Extract :class:`pyttb.sptensor` values as specified by `W`, a
+        :class:`pyttb.sptensor` containing only non-zero values that are
+        ones (1). The values in the :class:`pyttb.sptensor` corresponding
+        to the indices of the ones (1) in `W` will be returned as a column
```
ntjohnson1, line 1174: This might be a little clearer?
```suggestion
        Extract :class:`pyttb.sptensor` values as specified by `W`.
        The values in the :class:`pyttb.sptensor` corresponding
        to entries containing ones (1) in `W` will be returned as a column
```

pyttb/sptensor.py
```diff
@@ -860,32 +1232,56 @@ def mask(self, W: sptensor) -> np.ndarray:
...
         """
-        Matricized tensor times Khatri-Rao product for sparse tensor.
+        Matricized tensor times Khatri-Rao product for sparse tensor. This is
+        an efficient form of the matrix product that avoids explicitly
+        computing the matricized sparse tensor and the large intermediate
+        Khatri-Rao product arrays.
+
+        If the input includes a list of 2-D arrays (factor_matrices), this
+        computes a matrix product of the mode-`n` matricization of the sparse
+        tensor the with Khatri-Rao product of all arrays in the list except
```
ntjohnson1, line 1242:
```suggestion
        tensor with the Khatri-Rao product of all arrays in the list except
```

pyttb/sptensor.py
```diff
@@ -860,32 +1232,56 @@ def mask(self, W: sptensor) -> np.ndarray:
...
+        tensor the with Khatri-Rao product formed by the `factor_matrices` and
+        `weights` from the ktensor, excluding the `n`th factor matrix and
+        corresponding weight. The shape of the ktensor must be compatible with
+        the shape of the sparse tensor.
 
         Parameters
         ----------
         U:
-            Matrices to create the Khatri-Rao product
+            List of 2-D arrays or ktensor.
```
ntjohnson1, line 1258: Same conceptual note about our documentation adding contextual information beyond types. Since the types are already present and testable.

ntjohnson1 (commented):
I made it about halfway through the sptensor.py code changes. At `nvecs` currently. A few recurring notes came up so sharing at this point in case you have thoughts. Otherwise I should have time to circle back to the code level changes this weekend and still need to scroll top to bottom on the read the docs generated page to check for rendering changes.

pyttb/sumtensor.py
```diff
@@ -108,9 +108,9 @@ def __repr__(self):
             tensor of shape (2, 2)
             data[:, :] =
             [[1. 1.]
-            [1. 1.]]
+             [1. 1.]]
```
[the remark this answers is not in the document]
dmdunla: The NORMALIZE_WHITESPACE ignores tabs and spaces. Since you were only indenting with `\t` before each part, I fixed the indentation that `numpy` provides for the `tensor.data` objects. This now reflects the output where only the `\t` is normalized.

pyttb/sptensor.py
```diff
@@ -339,41 +422,52 @@ def from_aggregator(
...
         return ttb.sptensor(self.subs, self.vals, self.shape, copy=True)
 
     def __deepcopy__(self, memo):
         return self.copy()
 
-    # TODO decide if property
     def allsubs(self) -> np.ndarray:
         """
-        Generate all possible subscripts for sparse tensor
+        Generate all possible subscripts for sparse tensor.
```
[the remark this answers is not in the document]
dmdunla: MATLAB is for math people (subs/subscripts) and Python is for CS people (indices). I got lost in trying to resolve these issues. The only place in `numpy` where subscripts are mentioned are in `einsum`. I started moving away from subscripts, but I think it would be better to align with MATLAB for now, then make a decision post-v2 if we want subscripts or indices.

pyttb/sptensor.py
dmdunla: Agreed. WIll make the change.

pyttb/sptensor.py
dmdunla: Agreed, will make this change.

pyttb/sptensor.py
dmdunla: Agreed, will make this change.

pyttb/sptensor.py
```diff
@@ -120,6 +120,32 @@ def tt_from_sparse_matrix(
 class sptensor:
     """
     SPTENSOR Class for sparse tensors.
+
+    Contains the following data members:
+
+    ``subs``: `p` x `n` array (:class:`numpy.ndarray`) specifying the
```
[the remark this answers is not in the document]
dmdunla: Agreed, will make this change.

pyttb/sptensor.py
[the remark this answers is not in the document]
ntjohnson1: Sounds good. This is related to our previous discussion about making some note on unifying language/style. No change needed.

ntjohnson1 (approved):
This is everything I could find by a line review on the code and scrolling through the updated sptensor page on RTD. I think the only remaining thing besides the direct suggestions are the COMPARISON_ERROR and the changes to the logical_* methods. I'm fine if you want to merge this with those to filed as a follow up.

dmdunla:
I got through most of your changes and suggestions. However, sptensor is more streamlined than the other methods, in terms of docs. This is an evolving process, so I may revisit those to help clean up while I’m on this kick. But sptensor took way more time than the other classes, so we’ll see.

I hope to merge today at some point.
________________________________
From: Nick ***@***.***>
Sent: Saturday, November 4, 2023 10:05:48 AM
To: sandialabs/pyttb ***@***.***>
Cc: Dunlavy, Danny ***@***.***>; Assign ***@***.***>
Subject: [EXTERNAL] Re: [sandialabs/pyttb] 21 sptensor complete documentation for all methods (PR #219)


@ntjohnson1 approved this pull request.

This is everything I could find by a line review on the code and scrolling through the updated sptensor page on RTD. I think the only remaining thing besides the direct suggestions are the COMPARISON_ERROR and the changes to the logical_* methods. I'm fine if you want to merge this with those to filed as a follow up.

________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382277485>:

> @@ -1016,16 +1531,65 @@ def reshape(
         old_modes: Optional[Union[np.ndarray, int]] = None,
     ) -> sptensor:
         """
-        Reshape specified modes of sparse tensor
+        Reshape the sparse tenso to the have shape specified in `new_shape`.


⬇️ Suggested change

-        Reshape the sparse tenso to the have shape specified in `new_shape`.
+        Reshape the sparse tensor to the have shape specified in `new_shape`.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382277650>:

> @@ -1016,16 +1531,65 @@ def reshape(
         old_modes: Optional[Union[np.ndarray, int]] = None,
     ) -> sptensor:
         """
-        Reshape specified modes of sparse tensor
+        Reshape the sparse tenso to the have shape specified in `new_shape`.
+        If `old_modes` is specified, reshape only those modes, moving newly
+        reshape modes to the end of the indices. The product of the new shape


⬇️ Suggested change

-        reshape modes to the end of the indices. The product of the new shape
+        reshaped modes to the end of the indices. The product of the new shape


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382278274>:

>
         Parameters
         ----------
-        new_shape:
+        shape:


This doesn't match the name of the actual parameter (there's probably a ruff flag to enable checking this). Did you mean to update the param name to just be shape?

________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382279460>:

>          siz = tuple(shapeArray[idx])
         if self.vals.size == 0:
             return ttb.sptensor(np.array([]), np.array([]), siz, copy=False)
         return ttb.sptensor(self.subs[:, idx], self.vals, siz)

     def subdims(self, region: Sequence[Union[int, np.ndarray, slice]]) -> np.ndarray:
         """
-        SUBDIMS Compute the locations of subscripts within a subdimension.
+        Compute the locations of subscripts within a subdimension.
+
+        Finds the locations of the subscripts in the sparse tensor that are
+        within the range specified by `region`. For example, if `region` is
+        `[1, np.array([1,2]), np.array([1,2]])`, then the locations of all
+        elements of the sparse tensor that have a first index equal to 1,
+        a second index equal to 1 or 2, and a third index equal to 1 or 2
+        is returned.


⬇️ Suggested change

-        is returned.
+        are returned.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382280251>:

> +
+        Returns
+        -------


Looks like a bad copy paste. Returns is defined after Parameters here

⬇️ Suggested change

-
-        Returns
-        -------


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382280370>:

>
         Parameters
         ----------
         vector:
-            Vector(s) to multiply against
+            Vector or vectors to multiple by.


⬇️ Suggested change

-            Vector or vectors to multiple by.
+            Vector or vectors to multiply by.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382282900>:

>
         Parameters
         ----------
-        other:
-            Compare equality of sptensor to other
+        other: other: :class:`pyttb.sptensor`, float, int


So mypy gets upset when we type dunder methods because they should support generic objects. However we should be able to return NotImplemented for unsupported dunder methods and provide typing overloads that make both type checking and self documentation clearer. I'll file a ticket to explore that. For now I think having more type related documentation around the dunder methods is fine. (Especially to avoid any more scope creep on this PR)

⬇️ Suggested change

-        other: other: :class:`pyttb.sptensor`, float, int
+        other: :class:`pyttb.sptensor`, float, int


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382419425>:

>
         Parameters
         ----------
-        other
+        float, int


⬇️ Suggested change

-        float, int
+        other: float, int


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382419654>:

> @@ -2732,27 +3759,23 @@ def sptendiag(
     elements: np.ndarray, shape: Optional[Tuple[int, ...]] = None
 ) -> sptensor:
     """
-    Creates a sparse tensor with elements along super diagonal
-    If provided shape is too small the tensor will be enlarged to accomodate
+    Creates a sparse tensor with elements along super diagonal. If provided
+    shape is too small the sparse tensor will be enlarged to accomodate.


This typo existed before and was almost certainly introduced by me.

⬇️ Suggested change

-    shape is too small the sparse tensor will be enlarged to accomodate.
+    shape is too small the sparse tensor will be enlarged to accommodate.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382419957>:

> @@ -2373,19 +3322,41 @@ def __gt__(self, other):
             )

         # Otherwise
-        assert False, "Cannot compare sptensor with that type"
+        assert False, "Comparison allowed with sptensor, tensor, or scalar only."


NIT: If there are any changes that require actually editing code can we make this error message a global variable in sptensor?
We use it a handful of times. Not trying to scope creep too much so also open to this being future cleanup. If feeling really motivated could then import the global to the tests in case we change the common message in the future.

COMPARISON_ERROR = "Comparison allowed with sptensor, tensor, or scalar only."
...
assert False, COMPARISON_ERROR

________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382420778>:

> +        Sparse tensor with True set to all zero values from original sparse
+        tensor.


Right now this renders as two separate bullets. The continuation and indent makes it one line that wraps in sphinx.

⬇️ Suggested change

-        Sparse tensor with True set to all zero values from original sparse
-        tensor.
+        Sparse tensor with True set to all zero values from original sparse \
+           tensor.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421046>:

>          old_modes:
+            modes used for reshaping (default is all modes)


Globally I think you aligned on capitalizing the first letter and ending with period.

⬇️ Suggested change

-            modes used for reshaping (default is all modes)
+            Modes used for reshaping (default is all modes).


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421061>:

>
         Parameters
         ----------
-        new_shape:
+        shape:
+            New shape


⬇️ Suggested change

-            New shape
+            New shape.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421186>:

> +        Convert it to a sparse matrix (scipy.sparse.coo_matrix):
+        >>> M = S.spmatrix()


Without the blank line it all renders together as a comment

⬇️ Suggested change

-        Convert it to a sparse matrix (scipy.sparse.coo_matrix):
-        >>> M = S.spmatrix()
+        Convert it to a sparse matrix (scipy.sparse.coo_matrix):
+
+        >>> M = S.spmatrix()


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421534>:

> -        Returns
-        -------
-        Copy of current sparse tensor with empty slices removed.


This seems to be one of the only methods without a Returns. Was that intentional?

________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421643>:

>          shape:
-            Shape of generated tensor
+            Shape of generated sparse tensor


⬇️ Suggested change

-            Shape of generated sparse tensor
+            Shape of generated sparse tensor.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421653>:

>          nonzeros:
-            Number of nonzeros in generated tensor
+            Number of nonzeros in generated sparse tensor


⬇️ Suggested change

-            Number of nonzeros in generated sparse tensor
+            Number of nonzeros in generated sparse tensor.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421714>:

>          """
         Contract tensor along two dimensions (array trace).

         Parameters
         ----------
-        i:
-            First dimension
-        j:
-            Second dimension
+        i_0:
+            First dimension.
+        i_1:
+            Second dimension.

         Returns
         -------
         Contracted sptensor, converted to tensor if sufficiently dense


⬇️ Suggested change

-        Contracted sptensor, converted to tensor if sufficiently dense
+        Contracted sptensor, converted to tensor if sufficiently dense.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382421895>:

>          """
         return self.subs, self.vals

     def to_tensor(self) -> ttb.tensor:
-        """Convenience method to convert to tensor.
+        """
+        Convenience method to convert to tensor.
         Same as :meth:`pyttb.sptensor.full`


⬇️ Suggested change

-        Same as :meth:`pyttb.sptensor.full`
+        Same as :meth:`pyttb.sptensor.full`.


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382422065>:

> +
+        Returns
+        -------
+        Computed eigenvectors.
+
+        Examples
+        --------
+        Create a 2-way sparse tensor:
+
+        >>> subs = np.array([[0, 0], [0, 1], [1, 0]])
+        >>> vals = np.array([[1.0], [2.0], [3.0]])
+        >>> shape = (2, 2)
+        >>> S = ttb.sptensor(subs, vals, shape)
+
+        Compute two mode-0 leading eigenvectors, making sign of largest element
+        of each eigenvector positive (i.e., `flipsign=True, which is the


⬇️ Suggested change

-        of each eigenvector positive (i.e., `flipsign=True, which is the
+        of each eigenvector positive (i.e., `flipsign`=True, which is the


________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382422551>:

>      def allsubs(self) -> np.ndarray:
         """
-        Generate all possible subscripts for sparse tensor
+        Generate all possible subscripts for sparse tensor.


Sounds good. This is related to our previous discussion about making some note on unifying language/style. No change needed.

________________________________

In pyttb/sptensor.py<https://github.com/sandialabs/pyttb/pull/219#discussion_r1382420355>:

>
         Parameters
         ----------
-        other:
-            Compare equality of sptensor to other
+        other: other: :class:`pyttb.sptensor`, float, int


Ticket filed #288<https://github.com/sandialabs/pyttb/issues/288>

pyttb/sptensor.py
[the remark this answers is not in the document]
dmdunla: Oops, I made all of these changes to subscripts in my local branch already before seeing this update. I think I'll make the change and we can standardize with a follow-up issue.

pyttb/sptensor.py
dmdunla: Actually streamlined to just "Mask tensor." to align with brevity throughout.

pyttb/sptensor.py
```diff
@@ -1016,16 +1531,65 @@ def reshape(
...
+        Reshape the sparse tenso to the have shape specified in `new_shape`.
+        If `old_modes` is specified, reshape only those modes, moving newly
+        reshape modes to the end of the indices. The product of the new shape
+        must equal the product of the old shape, in both the default case and
+        when using `old_modes`.
 
         Parameters
         ----------
-        new_shape:
+        shape:
```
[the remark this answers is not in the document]
dmdunla: Nope, fixed now.

pyttb/sptensor.py
```diff
@@ -2373,19 +3322,41 @@ def __gt__(self, other):
             )
 
         # Otherwise
-        assert False, "Cannot compare sptensor with that type"
+        assert False, "Comparison allowed with sptensor, tensor, or scalar only."
```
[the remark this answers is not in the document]
dmdunla: I'm going to leave this for the future. If you want to address it, please add an issue. Just trying to avoid prolonging this already way overdue PR.

pyttb/sptensor.py
```diff
@@ -2651,15 +3665,29 @@ def squash(
...
+        >>> S2, inverse = S.squash(True)
+        >>> for i in range(S.ndims):
+        ...     np.array_equal(S.subs[:, i], inverse[i][S2.subs[:, i]])
+        True
+        True
         True
-
-        Returns
-        -------
-        Copy of current sparse tensor with empty slices removed.
```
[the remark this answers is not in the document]
dmdunla: I'm removing "Returns", as they are redundant with type hints. We should create an issue and decide what we want to do in the future. The formatting of the "Returns" was all over the place in `sptensor`.

PR #237, opened 2023-09-15
Default behavior of divmod() is different from MATLAB mod(), on which…

jeremy-myers:
… the original logic was based. The prior logic threw a ZeroDivisionError if printitn == 0. Instead, this fix avoids this error by testing that printitn > 0. Tests added for various values of printitn.

----
:books: Documentation preview :books:: https://pyttb--237.org.readthedocs.build/en/237/

### DIFF
--- a/tests/test_cp_als.py
+++ b/tests/test_cp_als.py
@@ -213,21 +213,17 @@ def test_cp_als_tensor_printitn(capsys, sample_tensor):
     _, T = sample_tensor
 
     # default printitn
-    _, _, output = ttb.cp_als(T, 2, printitn=1)
+    ttb.cp_als(T, 2, printitn=1, maxiters=2)
     capsys.readouterr()
-    assert pytest.approx(output["fit"]) == 1
-
+    
     # zero printitn
-    (M, Minit, output) = ttb.cp_als(T, 2, printitn=0)
+    ttb.cp_als(T, 2, printitn=0, maxiters=2)
     capsys.readouterr()
-    assert pytest.approx(output["fit"]) == 1
-
+    
     # negative printitn
-    (M, Minit, output) = ttb.cp_als(T, 2, printitn=-1)
+    ttb.cp_als(T, 2, printitn=-1, maxiters=2)
     capsys.readouterr()
-    assert pytest.approx(output["fit"]) == 1
 
     # float printitn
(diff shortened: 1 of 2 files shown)


tests/test_cp_als.py
```diff
@@ -206,3 +206,28 @@ def test_cp_als_sptensor_zeros(capsys):
     capsys.readouterr()
     assert pytest.approx(output3["fit"], 1) == 0
     assert output3["normresidual"] == 0
+
+
+@pytest.mark.indevelopment
```
ntjohnson1, line 211: No need for this mark. We don't use it anywhere and I think I've stripped them from a few of our test files. I should probably do a global remove at some point

tests/test_cp_als.py
```diff
@@ -206,3 +206,28 @@ def test_cp_als_sptensor_zeros(capsys):
...
+def test_cp_als_tensor_printitn(capsys, sample_tensor):
+    (data, T) = sample_tensor
```
ntjohnson1, line 213: Style note (you are following what is in this file, but I haven't revisited it recently so it should be cleaned up eventually), there is no need for the parenthesis when returning tuples. We also don't use data anywhere.
```suggestion
    _, T = sample_tensor
```

tests/test_cp_als.py
```diff
@@ -206,3 +206,28 @@ def test_cp_als_sptensor_zeros(capsys):
...
+
+    # default printitn
+    (M, Minit, output) = ttb.cp_als(T, 2, printitn=1)
```
ntjohnson1, line 216: Similar tuple note as above
```suggestion
    _, _, output = ttb.cp_als(T, 2, printitn=1)
```

tests/test_cp_als.py
```diff
@@ -206,3 +206,28 @@ def test_cp_als_sptensor_zeros(capsys):
...
+    capsys.readouterr()
+    assert pytest.approx(output["fit"]) == 1
```
ntjohnson1, line 218: We don't actually care about the fit here so can we ignore checking it? As far as I can tell this is just a smoke test.
In which case it might also be nice to set `maxiters` to something small to save the CPU cycles.

ntjohnson1 (commented):
A few small desired cleanups then looks good to me and @dmdunla can merge. Thanks!

ntjohnson1 (approved):

Issue #236, opened 2023-09-14
`cp_als` does not accept params tuple as input


jeremy-myers:
In MATLAB TTB, the third output argument from `cp_als` is a struct with a field `params` that can be used to re-start `cp_als` with the exact same configuration. In pyttb, the third output argument is a dict with a field `params`. However, due to the order in which the params are assigned vs. the order of the input arguments to `cp_als`, this causes a failure.

# Minimal working example
```
R = 3
np.random.seed(0)  # Set seed for reproducibility
X = ttb.tensor(np.random.rand(6,8,10), shape=(6,8,10))
M, _, info = ttb.cp_als(X,3,printitn=100)
M2 = ttb.cp_als(X,3,*info['params'])

---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[106], line 2
      1 M, _, info = ttb.cp_als(X,3,printitn=100)
----> 2 M2 = ttb.cp_als(X,3,*info['params'])

File ~/pyttb/pyttb/cp_als.py:133, in cp_als(input_tensor, rank, stoptol, maxiters, dimorder, init, printitn, fixsigns)
    131     dimorder = list(range(N))
    132 elif not isinstance(dimorder, list):
--> 133     assert False, "Dimorder must be a list"
    134 elif tuple(range(N)) != tuple(sorted(dimorder)):
    135     assert False, "Dimorder must be a list or permutation of range(tensor.ndims)"

AssertionError: Dimorder must be a list
```

I assume this is because the order of `info['params']` is `(stoptol, maxiters, printitn, dimorder)` but the order of the `cp_als` args is (after `rank`): `(stoptol, maxiters, dimorder, init, printitn)`. If these were aligned, this functionality would likely work.

PR #238, opened 2023-09-15
Changes output['params'] from tuple to dict, enhancing output express…

jeremy-myers:
…iveness and ability to pass output['params'] to cp_als() by unpacking operator **.

----
:books: Documentation preview :books:: https://pyttb--238.org.readthedocs.build/en/238/

PR #241, opened 2023-09-16
240 cp apr does not accept params tuple as input

jeremy-myers:
Fixes https://github.com/sandialabs/pyttb/issues/236#issue-1897495997 in a way similar to https://github.com/sandialabs/pyttb/pull/238#issue-1898957283. Also fixes a test in `test_cpapr_pqnr()` where an error on bad line search was thrown before the target of the test was evaluated.

----
:books: Documentation preview :books:: https://pyttb--241.org.readthedocs.build/en/241/


tests/test_cp_apr.py
```diff
@@ -238,12 +250,18 @@ def test_cpapr_pqnr(capsys):
...
     assert np.isclose(M.full().data, ktensorInstance.full().data, rtol=1e-01).all()
 
+    # Assert that params from previous run can be provided as input
+    M, _, output2 = ttb.cp_apr(sptensorInstance, 2, **output["params"])
+    capsys.readouterr()
+    assert output["params"] == output2["params"]
+
     # Edge cases
     # Confirm timeout works
+    np.random.seed(123)
```
ntjohnson1, line 264: This seed confirms reproducibility for a given machine but not across different machines. Does it make more sense to just dump an explicit initial guess we know doesn't have an error at the first step? Was the bad line search actually making this test irrelevant or just creating noise in the pytest output? Because it looks like we still captured the time limit exceeded which is the goal of the test anyway.

If its just reducing the print out can we just catch/squash the print. If the previous test was fundamentally broken and the tensor is small can we just use an explicit initial value that doesn't hit the line search issue?

tests/test_cp_apr.py
```diff
@@ -238,12 +250,18 @@ def test_cpapr_pqnr(capsys):
     M, _, _ = ttb.cp_apr(sptensorInstance, 2, algorithm="pqnr")
     capsys.readouterr()
     assert np.isclose(M.full().data, ktensorInstance.full().data, rtol=1e-01).all()
-    M, _, _ = ttb.cp_apr(sptensorInstance, 2, algorithm="pqnr", precompinds=False)
+    M, _, output = ttb.cp_apr(sptensorInstance, 2, algorithm="pqnr", precompinds=False)
     capsys.readouterr()
     assert np.isclose(M.full().data, ktensorInstance.full().data, rtol=1e-01).all()
 
+    # Assert that params from previous run can be provided as input
+    M, _, output2 = ttb.cp_apr(sptensorInstance, 2, **output["params"])
```
ntjohnson1, line 258: IIRC cp_apr is the slowest of our current tests. Globally, can we also set max iters to a low value throughout the newly added tests since we don't care about the result? In general our test suite runs locally pretty quickly, but trying to prevent gradual bloat.

ntjohnson1 (commented):
Thanks! Two minor notes but otherwise looks good.

tests/test_cp_apr.py
dmdunla: Suggestion: using an initial guess of all ones (weights and factor_matrices) is a typical choice for deterministic testing.

tests/test_cp_apr.py
dmdunla: Suggestion: use 10 for the maximum number of iterations

dmdunla: Or at least enough to make it through a single pass over all modes/dimensions + 1. This will make sure to test outer iteration and inneriteration.

tests/test_cp_apr.py
jeremy-myers: I added fixtures for two tensor problems with their solutions as well as two initial guess ktensors. Finding the right combo of problems/initial guesses that ensured PQNR threw a bad L-BFGS error made for a somewhat ugly solution (i.e., 2 tensor inputs & 2 initial guesses), but this ensures that functionality is flexed.

tests/test_cp_apr.py
jeremy-myers: IMO this probably warrants its own issue, so I'll open a new issue for it.

dmdunla (approved):

PR #254, opened 2023-09-20
252 gcp opt valueerror setting an array element with a sequence

jeremy-myers:
----
:books: Documentation preview :books:: https://pyttb--254.org.readthedocs.build/en/254/


ntjohnson1:
FYI I would prefer the solution here https://github.com/sandialabs/pyttb/pull/255 over the try/except version.

ntjohnson1:
Duplicates https://github.com/sandialabs/pyttb/pull/255

Issue #258, opened 2023-09-25
`sptensor.mask` not working as intended 


dmdunla:
The `sptensor.mask` method does not work as intended when one of the non-zeros in the mask corresponds to a zero in the orignal sparse tensor. This is not tested in `tests/test_sptensor.py`.

```python
>>> import pyttb as ttb
>>> import numpy as np

>>> S = ttb.sptensor()
>>> S[0,0] = 1; S[1,1] = 2
>>> S
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = 1.0
[1, 1] = 2.0

>>> W = ttb.sptensor()
>>> W[0,0] = 1; W[1,0] = 1;
>>> W
sparse tensor of shape (2, 1) with 2 nonzeros
[0, 0] = 1.0
[1, 0] = 1.0

>>> S.mask(W) # this is incorrect
array([[1.],
       [2.]])

>>> W = ttb.sptensor()
>>> W[0,0] = 1; W[1,1] = 1;
>>> W
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = 1.0
[1, 1] = 1.0

>>> S.mask(W) # this is correct
array([[1.],
       [2.]])

```

PR #259, opened 2023-09-25
Sptensor mask

ntjohnson1:
We were using `tt_ismember_rows` incorrectly in mask. `-1` was a flag for failed match but also yields a valid index into an array. First fixed the bug then updated the ismember_rows interface to try to make this incorrect usage more obvious in the future.

Resolves https://github.com/sandialabs/pyttb/issues/258

----
:books: Documentation preview :books:: https://pyttb--259.org.readthedocs.build/en/259/


dmdunla (approved):
Good catch!

Issue #257, opened 2023-09-25
`sptensor.logical_or` has inconsistent output


dmdunla:
As in #256, `sptensor.logical_or` has inconsistent output when the input is a dense tensor, scalar, and sparse tensor: 

```python
>>> import pyttb as ttb
>>> import numpy as np
>>> S = ttb.sptensor()
>>> S[0,0] = 1; S[1,1] = 2
>>> S
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = 1.0
[1, 1] = 2.0

>>> S.logical_or(S)
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = 1.0
[1, 1] = 1.0

>>> T = S.to_tensor()
>>> S.logical_or(T)
tensor of shape (2, 2)
data[:, :] =
[[ True False]
 [False  True]]

>>> S.logical_or(1)
tensor of shape (2, 2)
data[:, :] =
[[ True  True]
 [ True  True]]
```

PR #269, opened 2023-09-27
Sparse logical

### LINKED ISSUE #256: `sptensor.logical_and` does not have expected behavior
`sptensor.logical_and` behavior is different than in MATLAB, and I am not sure that what the MATLAB version is doing is useful/correct.

**Questions:** Do we want to match the MATLAB behavior? Do we want to follow `numpy`/`scipy.sparse`? If so, what are those behaviors.

Here is the current `sptensor.logical_and`:

```python
Create a sparse tensor:

>>> S = ttb.sptensor()
>>> S[0,0] = 1; S[1,1] = 2
>>> S
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = 1.0
[1, 1] = 2.0

Compute logical AND with dense tensor that has the same non-zero
pattern but different values:

>>> T = S.to_tensor()
>>> T[0,0] = T[0,0] + 1
>>> S.logical_and(T)
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = 1.0
[1, 1] = 1.0

Compute logical AND with scalar value:

>>> S.logical_and(1.0)
sparse tensor of shape (2, 2) with 2 nonzeros
[0, 0] = True
[1, 1] = False
```

The output when computing with a tensor is an indicator sparse tensor, whereas it is a boolean sparse tensor for the scalar case. Also, the scalar case is computing the AND as an equality comparison it appears, not just a check for non-zero.

Here is the output from MATLAB:

```matlab
>> S = sptensor()
S is an all-zero sparse tensor of size [empty tensor]
>> S(1,1) = 1
S is a sparse tensor of size 1 x 1 with 1 nonzeros
	(1,1)     1
>> S(2,2) = 2
S is a sparse tensor of size 2 x 2 with 2 nonzeros
	(1,1)     1
	(2,2)     2
>> T = tensor(S)
T is a tensor of size 2 x 2
	T(:,:) = 
	     1     0
	     0     2
>> T(1,1) = 2
T is a tensor of size 2 x 2
	T(:,:) = 
	     2     0
	     0     2
>> and(S,tensor(S))
ans is a sparse tensor of size 2 x 2 with 2 nonzeros
	(1,1)   1
	(2,2)   1
>> and(S,1.0)
ans is a sparse tensor of size 2 x 2 with 2 nonzeros
	(1,1)     1
	(2,2)     1

```

The scalar AND is odd in MATLAB, as it checks for either zero or non-zero depending on the scalar input:

```matlab
>> and(S,0)
ans is an all-zero sparse tensor of size 2 x 2
>> and(S,1)
ans is a sparse tensor of size 2 x 2 with 2 nonzeros
	(1,1)     1
	(2,2)     1
>> and(S,2)
ans is a sparse tensor of size 2 x 2 with 2 nonzeros
	(1,1)     1
	(2,2)     1
```

The documentation in `pyttb` should document this behavior (once it matches with MATLAB).

Here are some experiments with `numpy` and `scipy`, where I am unable to compute `__and__` at all, so I am unsure what the default behavior for those should be:

```python
>>> import numpy as np
>>> A = np.zeros((2,2))
>>> A[0,0] = 1
>>> A & 0
TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
>>> 0 & A
TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

>>> import scipy.sparse as sp 
>>> B = sp.coo_matrix(A)
>>> B
<2x2 sparse matrix of type '<class 'numpy.float64'>'
        with 1 stored elements in COOrdinate format>
>>> B & 1
TypeError: unsupported operand type(s) for &: 'coo_matrix' and 'int'
>>> B & A
TypeError: unsupported operand type(s) for &: 'coo_matrix' and 'float'
>>> B.toarray()
array([[1., 0.],
       [0., 0.]])
>>> B.toarray() & A
TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

ntjohnson1:
Resolves #256 and #257 but maybe raises more questions than it answers. It matches what you described as expected behavior. (Indicator tensors with 1.0, 0.0)

It raises the question on if we should track the dtype (could be a hidden variable or a property). Or even more explicitly could be set when constructing the tensor (I kind of hacked the vals in the sptensor to make things make sense; however would be nice to take a dtype argument on construction).

You mentioned in the issue we don't do boolean tensors but for equality we currently do. Should those also get the same indicator treatment?
```python
>> import pyttb as ttb
>> T = ttb.tenones((2,2))
>> T == T
tensor of shape (2, 2)
data[:, :] =
[[ True  True]
 [ True  True]]
```

----
:books: Documentation preview :books:: https://pyttb--269.org.readthedocs.build/en/269/

### DIFF
--- a/pyttb/sptensor.py
+++ b/pyttb/sptensor.py
@@ -704,7 +704,7 @@ class sptensor:
                 C = sptensor(shape=self.shape)
             else:
                 newvals = self.vals == B
-                C = sptensor(self.subs, newvals, self.shape)
+                C = sptensor(self.subs, newvals.astype(self.vals.dtype), self.shape)
             return C
         # Case 2: Argument is a tensor of some sort
         if isinstance(B, sptensor):
@@ -718,6 +718,7 @@ class sptensor:
                 self.shape,
                 lambda x: len(x) == 2,
             )
+            C.vals = C.vals.astype(self.vals.dtype)
 
             return C
 
@@ -741,7 +742,7 @@ class sptensor:
         allsubs = self.allsubs()
         subsIdx = tt_setdiff_rows(allsubs, self.subs)
         subs = allsubs[subsIdx]
-        trueVector = np.ones(shape=(subs.shape[0], 1), dtype=bool)
+        trueVector = np.ones(shape=(subs.shape[0], 1), dtype=self.vals.dtype)
         return sptensor(subs, trueVector, self.shape)
 
     @overload
@@ -771,12 +772,14 @@ class sptensor:
             assert False, "Logical Or requires tensors of the same size"
 
         if isinstance(B, ttb.sptensor):
-            return sptensor.from_aggregator(
+            C = sptensor.from_aggregator(
                 np.vstack((self.subs, B.subs)),
                 np.ones((self.subs.shape[0] + B.subs.shape[0], 1)),
                 self.shape,
                 lambda x: len(x) >= 1,
             )
+            C.vals = C.vals.astype(self.vals.dtype)
+            return C
 
         assert False, "Sptensor Logical Or argument must be scalar or sptensor"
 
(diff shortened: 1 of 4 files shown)


dmdunla (approved):

dmdunla:
Approved, but the bigger issue is explicitly handling `dtype`s for our tensor classes.

ntjohnson1:
Do you mind filing a ticket for that issue just so we don't forget? Not sure if it's needed for 2.0 or not

Issue #208, opened 2023-08-10
Add nbstripout to dev dependencies and commit checks


dmdunla:
The following is useful for keeping the jupyter notebook tutorials clean (output and metadata removed):

```
nbstripout docs/source/tutorial/class_ttensor.ipynb \
--extra-keys "metadata.language_info metadata.vscode metadata.kernelspec"
```

This will require including `nbstripout` into the dev dependencies and adding a hook for pre-commits.

ntjohnson1:
I looked at adding a check for this, but ran into an issue where nbstripout's check option would fail because it would want to edit something but the change was only the windows line endings not tracked by git. There's hopefully a work around for that but I didn't have time to dig deeper

ntjohnson1:
This is much more complicated that desired.
The dry-run check on nbstripout is kind of broken currently: https://github.com/kynan/nbstripout/issues/158 we could probably put in the leg work to fix it, but I don't have immediate bandwidth for that.
However, it looks like there is a github action to basically just run our pre-commit hooks on a repo https://github.com/pre-commit/action NICE!
Even better there is an action that could run the pre-commit hook and commit the changes https://pre-commit.ci/lite but that requires permissions that probably aren't acceptable (or at least would need to check). This would be nice because it would fix the times we forget to run black etc locally or doing a merge on the gui.

PR #276, opened 2023-09-28
Tutorial fixes

jeremy-myers:
Should fail in regression tests but we need to determine how it fails.

----
:books: Documentation preview :books:: https://pyttb--276.org.readthedocs.build/en/276/


jeremy-myers:
Changing the title of this PR since it was an easy fix and covers a section on plotting in the ttensor tutorial.

docs/source/tutorial/class_ttensor.ipynb
```diff
@@ -654,12 +822,56 @@
...
+    "axs[1, 0].set_title(\"Partial Reconstruction\")\n",
+    "axs[1, 1].set_title(\"Partial+Downsampled Reconstruction\")\n",
+    "\n",
+    "axs[1, 1].set_xlim = axs[1, 0].get_xlim()\n",
+    "axs[1, 1].set_ylim = axs[1, 0].get_ylim()"
+   ]
   }
  ],
- "metadata": {},
+ "metadata": {
```
dmdunla, line 868: Run the following before finalizing (from #208 which has not yet landed):

```
nbstripout docs/source/tutorial/class_ttensor.ipynb \
--extra-keys "metadata.language_info metadata.vscode metadata.kernelspec"
```

dmdunla (changes requested):
Please run nbstripout to remove the metadata

docs/source/tutorial/algorithm_cp_apr.ipynb
```diff
@@ -17,20 +17,18 @@
...
    ]
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "vscode": {
+     "languageId": "python"
+    }
```
dmdunla, line 30: Adding an addition `nbstripout` key to remove, as I had never seen these before (I don't use vscode for notebooks). The following will run on all tutorial files and fix all occurrences of the vscode metadata in the cells:

```
nbstripout docs/source/tutorial/*.ipynb --extra-keys "metadata.language_info metadata.vscode metadata.kernelspec cell.metadata.vscode"
```

dmdunla (changes requested):
Update `nbstripout` call to clean unwanted metadata.

jeremy-myers:
Resolves #277

dmdunla:
@ntjohnson1 Is the the expected output from your recent PR?

- #279 

It looks like the same file, docs/source/tutorial/algorithm_gcp_opt.ipynb, that needs to be updated. If so, do we just run `nbstripout` locally on this branch? Is it just the automatic `nbstripout` pre-commit locally that does not fully work ATM?

ntjohnson1:
> It looks like the same file, docs/source/tutorial/algorithm_gcp_opt.ipynb, that needs to be updated.

The error log doesn't mention that file, so I'm a little confused.

>  If so, do we just run nbstripout locally on this branch?

The pre-commit error message proposes one resolution `
If you are seeing this message in CI, reproduce locally with: 'pre-commit run --all-files'.` Or alternatively just running the nbstripout command in isolation should work (that's the instruction I added to the contributors doc).

>  Is it just the automatic nbstripout pre-commit locally that does not fully work ATM?

Nbstripout locally should work. On the issue the thing I pointed out is that the github action doesn't automatically fix this for us. We could set it up where if that test fails it just adds a commit that checks in the changes to unblock. But as far as I am aware right now everything works and is enforced.

dmdunla:
@ntjohnson1 Oops, my mistake, the file in the failed CI test was: `/docs/source/tutorial/algorithm_cp_apr.ipynb`. That must be from an update in this PR. So, then the `nbstripout` is working.

I missed that suggestion and just focused on the error itself.  I think this will be good moving forward.

@jeremy-myers Can you look at the failed CI test and run the suggested code locally? Then commit the changes and I'll review.

ntjohnson1:
Ya no problem. I just wanted to make sure we were on the same page. All the automation is supposed to reduce the things we have to consciously check for (and I saw you had been manually guarding the notebook metadata). If it becomes a headache we can explore other options.

ntjohnson1:
Pre-commit has been optional but supported all along. I added a note to the contributors docs that if people don't want to use pre-commit they can also just run the nbstripout command defined there.

Nbstripout doesn't have the option to be configured by project.toml (AFAICT) so we can't just tell people to run it from root similar to black/ruff. Nbstripout's dry_run functionality also always says there is a change even if it was just run. Since we don't anticipate changing the tutorials THAT often I figure it being slightly less smooth is ok. But ideally it is still clear.

dmdunla (approved):

Issue #180, opened 2023-07-06
Create mapping of pyttb-TTB functionality


dmdunla:
To assist transitions from the Tensor Toolbox for MATLAB to pyttb, create some reference documentation of the mapping of methods and any important differences.

ntjohnson1:
I believe https://github.com/sandialabs/pyttb/wiki/Development-Plans already has moderate coverage here (in case this wasn't already what you were thinking about).

dmdunla:
Yes, I was going to start from there (which is incomplete) and then consider moving it to RTD as a new section.

We're past the original intent of dev plans, but this will provide the basis for the MATLAB->pyttb mappings.

dmdunla:
This is related to:

- #110

As part of the RTD docs, I suggest we add a subsection of Getting Started called, "pyttb for MATLAB Users"

ntjohnson1:
I saw @jeremy-myers started to work on this already by writing up tables in restructured text (which looks great so far!). I would really like something that fails loudly when we continue to updated pyttb and potentially diverge from the MATLAB. We could probably setup the MATLAB getting started stuff as Jupyter notebooks rendered to our read the docs as well. Then if we change and interface that breaks it won't be allowed to check in. There are probably other ways.

Alternatively I think we can just add doctests to sphinx/RST if we think that would be easier. I'm not sure if the ability to interact with a jupyter notebook is as relevant for users coming from MATLAB.

jeremy-myers:
I've been quiet the last few weeks but expect to see some commits ones this PR this upcoming week.

ntjohnson1:
Zero rush on my side. I was on a variety of flights this week so I was much more active. Just wanted to weigh in to reduce a large refactor after this was 100% complete.

dmdunla:
Closed with 42f12892bae58bea3bf6a37a8dd12f31da2f0ec5

PR #291, opened 2023-12-12
Mapping of pyttb ttb functionality

jeremy-myers:
Addresses https://github.com/sandialabs/pyttb/issues/180

Summary: this PR came from a desire to ease the transition for MATLAB TTB users to pyttb. The main goal of mapping TTB and pyttb functionality is to document where naming conventions, method and data member usages, etc. differ between implementations. 

----
📚 Documentation preview 📚: https://pyttb--291.org.readthedocs.build/en/291/


jeremy-myers:
This PR fails regression tests, but only documentation was updated (no source code).

ntjohnson1:
It looks like there was a breaking change in ruff. Could try `pip install --upgrade ruff` which should repro locally. Then the failure says it an auto fixable change so `ruff check --fix .` from root should resolve. We probably should pin our ruff version.

jeremy-myers:
My pre-commit failed to update ruff due to proxy issues and missed this error. I'm not sure how to resolve these extra args calls in `cp_apr` and `fg_est`.

ntjohnson1:
This is a general configuration issue. Unrelated to your change like you mentioned. It looks like pre-commit requires (or I just set it up this way), a pinned version of ruff [here](https://github.com/sandialabs/pyttb/blob/5751f02f631a78dcf0a06ec4f03dbda79eb5b98b/.pre-commit-config.yaml#L3) that hasn't conflicted with latest ruff until now. The simplest unblocking solution is probably to update our package ruff requirement [here](https://github.com/sandialabs/pyttb/blob/5751f02f631a78dcf0a06ec4f03dbda79eb5b98b/pyproject.toml#L39) to match exactly. This will require rolling back your latest commit and downgrading ruff locally.

Then can file a ticket to handle this cleaner. Ideally linking the two configs and not pinning to an EXACT ruff version.

ntjohnson1:
If that doesn't work I can debug it after business hours today on a separate branch.

jeremy-myers:
@ntjohnson1 It sounds like I should wait for the pre-commit yaml and/or pyproject to be updated to unblock?

ntjohnson1:
Confirmed [this](https://github.com/sandialabs/pyttb/pull/292) unblocks and I think it is a reasonable temporary or even medium term solution since its only a dev requirement.

IMO I think the easiest approach is to cherry-pick the [commit](c233907341550c6d63c4a903dcbc33b4651e558d) from there or just apply the change yourself and revert 75c0f1b6ffed2b7208cb7af8c1106d35468aa429. No need for a separate PR to unblock.

jeremy-myers:
@ntjohnson1 It looks like this fix is now failing with `black --check .`. Is this another pinning issue?

ntjohnson1:
Boo. I assume so since it looks like we have it pinned here https://github.com/sandialabs/pyttb/blob/5751f02f631a78dcf0a06ec4f03dbda79eb5b98b/.pre-commit-config.yaml#L8. Should be quick to test if it fixes the issue.

pyproject.toml
```diff
@@ -36,7 +36,7 @@ dev = [
     "nbstripout",
     "pytest",
     "pytest-cov",
-    "ruff",
+    "ruff==0.0.284",
```
ntjohnson1, line 39: You updated the pre-commit config but left this pinned. So we still have the issue of incompatible pre-commit and project. My suggestion is to be more explicit in our project so the alignment is guaranteed then figure out an easier way to keep them aligned in the future as a follow up

jeremy-myers: I think I have it. I pinned ruff but pre-commit autoupdate gave a conflicting version. So, `pre-commit autoupdate --repo https://github.com/psf/black` to update black only should do the trick.

ntjohnson1: The tests pass now but we still have the possibility of this pre-commit/black mismatch occurring again. Can you pin the version of black in this file to match the one in the pre-commit. We can figure out if there is an automated way to align these as follow up. In my opinion it would be nicer to keep the version of black that was originally in in the pre-commit and just pin it here then roll back the changes from the newer version of black. That way the content changes and formatting churned aren't coupled together.

docs/source/matlab/common.rst
```diff
@@ -0,0 +1,83 @@
...
++-----------------+----------------------+------------------------------------------------------------------------+
+
+Methods
+^^^^^^^
++-----------------+----------------------+------------------------------------------------------------------------+
+| MATLAB name     | ``pyttb`` name       | Calling convention                                                     |
++=================+======================+========================================================================+
+| ``and``         | ``logical_and``      | ``X.logical_and(Y)``                                                   |
++-----------------+----------------------+------------------------------------------------------------------------+
+| ``disp``        | ``__str__``          | ``X``                                                                  |
```
ntjohnson1, line 19: Hmm, for all the names `pyttb names` that include dunder methods I wonder if it makes sense to use their public equivalents. Technically they are all hidden methods that get exposed via some other hook. For example `__str__` and `__repr__` both get called under the hood for `str(X)` and `repr(X)`. For someone totally unaware of python just trying to directly translate from matlab based on this doc I'd really hope they don't end up with a bunch of `__str__`, `__rmul__` calls in their code

docs/source/matlab/common.rst
```diff
@@ -0,0 +1,83 @@
...
+| ``times``       | ``__mul__``          | ``X * Y``                                                              |
++-----------------+----------------------+------------------------------------------------------------------------+
+| ``uminus``      | ``__neg__``          | ``-X``                                                                 |
++-----------------+----------------------+------------------------------------------------------------------------+
+| ``uplus``       | ``__pos__``          | ``+X``                                                                 |
++-----------------+----------------------+------------------------------------------------------------------------+
+| ``xor``         | ``logical_xor``      | ``X.logical_xor(Y)``                                                   |
++-----------------+----------------------+------------------------------------------------------------------------+
+
+Copying
```
ntjohnson1, line 70: It might be useful to note this is fairly common for non-builtin types in python and include a link. Hereish? https://docs.python.org/3/library/copy.html

docs/source/matlab/symktensor.rst
```diff
@@ -0,0 +1,8 @@
+``symktensor``
```
ntjohnson1, line 1: IMO I think it might be clearer to just remove the placeholders. I'm not sure what the plan is for actually covering these other classes.

docs/source/matlab/tenmat.rst
```diff
@@ -0,0 +1,22 @@
+``tenmat``
```
ntjohnson1, line 1: I have a PR up now that updates the tenmat constructors. I believe this is correct currently but once Danny merges my PR this will go stale.

docs/source/for_matlab_users.rst
```diff
@@ -0,0 +1,14 @@
+pyttb for MATLAB users
+======================
+
+Already familiar with MATLAB Tensor Toolbox? To assist transitions from the
+Tensor Toolbox for MATLAB to pyttb, this guide documents some key differences.
```
ntjohnson1, line 5: It might be good to mention that the tutorials are similar to the MATLAB ones if they want a hands on approach. I think that's already implied but might not be a bad idea to be explicit.

ntjohnson1 (commented):
Will leave for Danny to give the official approval and merge it in. Thanks for putting this together.

A few notes:
* My preference for all the formatting headache would be to pin black in pyproject to the previous version that is on main, rollback the changes in *.ipynb and *.py.
* As I mentioned in the tenmat doc some of these things may go stale and there isn't currently a way to check if the pyttb side is broken. Could you file a follow up ticket to add some sort of automated checking for this content? Either using Jupyter notebooks and the rendering we already have or sphinx supports doctest, but I don't think we have an example of that in the repo yet. Wouldn't need to verify outputs but just a smoke test basically to make sure the methods/functions actually exist.

dmdunla (approved):

PR #296, opened 2024-05-20
Add visualization for ktensor

ghbrown:
This PR adds the `vis` method to ktensor objects.
The syntax and options are very similar to the [TTB equivalent](https://gitlab.com/tensors/tensor_toolbox/-/blob/dev/@ktensor/viz.m?ref_type=heads) with one major change.

**No support for default plots**: the user must provide their own plotting function for each mode, so no calling like `K.vis()`.
**Why**: in practice users will probably provide their own custom plot functions anyway; they know more than us about their data and they likely won't like the "ugly" plots we would generate. If they want the default behavior it's +3 short lines of code for them (see the first example in docs).

An example showing parity with TTB for [this example](https://www.tensortoolbox.org/cp_als_doc.html):
image

To do:
- [ ] tests?

### DIFF
--- a/pyttb/ktensor.py
+++ b/pyttb/ktensor.py
@@ -20,6 +20,7 @@ from typing import (
     overload,
 )
 
+import matplotlib.pyplot as plt
 import numpy as np
 import scipy.sparse.linalg
 from typing_extensions import Self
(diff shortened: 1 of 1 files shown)


ghbrown:
I've added default behavior per a conversation with @dmdunla.

PR #302, opened 2024-06-26
Optimize only specific modes in cp_als()

ghbrown:
An argument called `optdims` of type `List[int]` is added to `cp_als()` to enable optimizing only the factor matrices of specific modes.
This was not previously possible even by scripting around `cp_als()` since it checks for convergence only after each full sweep.
Control flow was hardly altered (`dimorder` may now just be missing some entries), and the default is of course optimizing all modes.

----
📚 Documentation preview 📚: https://pyttb--302.org.readthedocs.build/en/302/

### DIFF
--- a/pyttb/cp_als.py
+++ b/pyttb/cp_als.py
@@ -198,7 +198,7 @@ def cp_als(  # noqa: PLR0912,PLR0913,PLR0915
     for iteration in range(maxiters):
         fitold = fit
 
-        # Iterate over all N modes of the tensor
+        # Iterate over optimized modes in specified order
         for n in dimorder:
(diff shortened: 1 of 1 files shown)


dmdunla:
@ghbrown Can you please add an issue as described in the GitHub workflow in [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md#github-workflow)? And please include a motivating use case for this so that is captured in the Issue.

ghbrown:
Thanks for the warning, Tammy. I just looked over that part of the code, and because of where and how I updated `dimorder` I believe all of that should be properly taken care of automatically.
But I did catch and fix a comment that needed to change!

PR #314, opened 2024-09-04
313 add timings for gcp opt with l bfgs

jeremy-myers:
* Adds a default callback to L-BFGS-B that tracks the times of each iteration. 
    * User-defined callbacks are set as a private member callable in a subclass to `LBFGSB` called `Monitor`
    * Adds tests to `tests/gcp/test_optimizers.py`
    * Aligns with MATLAB usage
    * Changes `monotonic()` to `perf_counter()` in `StochasticSolver`
* Adds timers to `gcp_opt` in `info` dictionary

----
📚 Documentation preview 📚: https://pyttb--314.org.readthedocs.build/en/314/


tests/gcp/test_optimizers.py
```diff
@@ -100,3 +100,33 @@ def test_lbfgsb(generate_problem):
...
+
+def test_lbfgsb_callback(generate_problem):
+    dense_data, model, sampler = generate_problem
+
+    # Test default callback
+    maxiter = 2
+    solver = LBFGSB(maxiter=maxiter)
+    _, info = solver.solve(model, dense_data, gaussian, gaussian_grad)
+    assert not "callback" in info["callback"].keys()
+    assert info["callback"]["time_trace"].shape == (maxiter,)
```
ntjohnson1, line 113: Can we update this to ensure the timings are greater than zero? When looking at your typing bug I accidentally turned off the timing portion but the member in the Monitor had the shape set correctly (but always returned zero)

tests/gcp/test_optimizers.py
```diff
@@ -100,3 +100,33 @@ def test_lbfgsb(generate_problem):
...
     assert not model.isequal(result)
     assert (model.full() - dense_data).norm() > (result.full() - dense_data).norm()
+
+
+def test_lbfgsb_callback(generate_problem):
+    dense_data, model, sampler = generate_problem
+
+    # Test default callback
+    maxiter = 2
+    solver = LBFGSB(maxiter=maxiter)
```
ntjohnson1, line 110: The code has a comment about re-using the solver but we don't test for it. Can we add a test to make sure that is supported correctly? (My MR to fix the types also changes how we reset things when we are done so a test would confirm our intentions align)

ntjohnson1 (approved):
LGTM. I added a PR into your branch [here](https://github.com/jeremy-myers/pyttb/pull/2) that is one approach to resolving the typing issues specific to this PR.

The other typing issues for python > 3.8 should be resolved [here](https://github.com/sandialabs/pyttb/pull/307) since they are partially related to numpy 2.0 and newer mypy. I didn't check them exhaustively so its possible that there is still some subtle error remaining but resolving those two PRs first should at least highlight what remains before digging too much deeper.

tests/gcp/test_optimizers.py
ntjohnson1: Yes. That's the request. Right now if ONLY the constructor for the monitor existed this test would still pass.

jeremy-myers: Ah okay, that makes sense.

jeremy-myers:
Regression tests fail on other code that was already failing, probably related to numpy & mypy in [this PR](https://github.com/sandialabs/pyttb/pull/307).

PR #326, opened 2024-10-25
Removing numpy pin.

dmdunla:
----
📚 Documentation preview 📚: https://pyttb--326.org.readthedocs.build/en/326/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.8"
 
 dependencies = [
-    "numpy < 2",
+    "numpy",
     "numpy_groupies",
     "scipy",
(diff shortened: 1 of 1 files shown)


ntjohnson1:
In #307 I put together a little utility `np_to_python` that I think resolves this pretty printing issue. The errors snuck in when we pinned numpy then added the tenant docs that came about before the change. I can probably clean this up this weekend. Or you can apply the utility to the few places.

dmdunla:
I'll let you take this one, since you had fixed some of the prior issues with the numpy scalar outputs.

ntjohnson1:
I don't have permission (or my remotes are messed up) to edit branches on mainline pyttb. Closing here and opening #329 to resolve

PR #330, opened 2024-11-03
Dev deps

ntjohnson1:
Based on some discussion after the pause in development.
1. Updates our pinned ruff
2. Replaces black with ruff format (faster and one less dependency)
3. Runs ruff format which formatted the code in our doc strings
4. Sets some versioning around our packages and uses UV as a pretty simple way to test the oldest version specified by our package.

There's definitely some future work that could be done on the versions if we wanted to be more diligent. I mostly just got it working. We used to support back to numpy 1.19 but we added a handful of features from 1.24. If there's a desire to support older deps we should be able to use the current setup to make the version width more flexible.

----
📚 Documentation preview 📚: https://pyttb--330.org.readthedocs.build/en/330/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.9"
 
 dependencies = [
-    "numpy",
+    "numpy<3.0,>=1.24",
     "numpy_groupies",
     "scipy",
     "matplotlib",
@@ -33,9 +33,11 @@ documentation = "https://pyttb.readthedocs.io"
 [project.optional-dependencies]
 dev = [
     "mypy<1.14.0",
-    "nbstripout",
+    # Also in pre-commit
+    "nbstripout<0.9",
     "pytest",
     "pytest-cov",
+    # Also in pre-commit
     "ruff<0.8",
     "pre-commit",
 ]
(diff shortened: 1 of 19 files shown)


pyproject.toml
```diff
@@ -32,18 +32,19 @@ documentation = "https://pyttb.readthedocs.io"
...
-    "mypy",
-    "nbstripout",
-    "pytest",
-    "pytest-cov",
-    "ruff==0.0.284",
-    "pre-commit",
+    "mypy>=1.10,<1.14.0",
+    # Also in pre-commit
+    "nbstripout>=0.8,<0.9",
+    "pytest>8.0",
```
dmdunla, line 38: Did something change in a recent version of pytest? In the past running `pytest .` would run all doctest and `tests/test_*.py` tests. Now, I need to run `pytest . tests` to get the `tests/test_*.py` tests to run. It looks like this was updated in the regression GH Action, but not in the docs.

dmdunla (changes requested):
All good to go except suggestion for a change in the info for contributors when running tests. The issue may have been introduced before, but I just noticed it with this PR.

pyproject.toml
ntjohnson1: I'm not sure. My history on my command line had the tests directory specified. I'll update the doc.

ntjohnson1: According to the pytest docs I'd expect it to still work unless I broke something in our config. I updated the doc but can check more later

dmdunla: Looks good.

dmdunla (approved):

PR #348, opened 2024-11-16
Nick/shape support

### LINKED ISSUE #305: Sptensor constructor breaks with int overflow for large tensor
# Issue
The constructor for `sptensor` fails an assertion when the tensor size is large.

# Example
For `size = (50000000, 50000000, 100, 50000)` and `nnzs = 1000000000`, we get:
```
    X = ttb.sptenrand(sz, nonzeros=nnz)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jermyer/pyttb/pyttb/sptensor.py", line 3709, in sptenrand
    return ttb.sptensor.from_function(unit_uniform, shape, valid_nonzeros)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jermyer/pyttb/pyttb/sptensor.py", line 214, in from_function
    assert False, (
AssertionError: Requested number of nonzeros must be positive and less than the total size
```

Checking `np.prod(size)` used in the assertion, `np.prod(size) = -6892481975075995648` .

# Solution
Use `prod` from the `math` module to get the correct result, i.e.,
```
>>> from math import prod
>>> sz = (50000000,50000000,100,50000)
>>> prod(sz)
12500000000000000000000
```
This has the added bonus that `sz` is a python `tuple`, not a `numpy` object, as pointed out by @dmdunla.

ntjohnson1:
This should make our shape arguments more flexible.
This provides machinery for simplifying when we want a vector (or scalar) of items. I didn't distinguish between an index array or a data array so technically this makes our typing slightly less specific. If that causes immediate concern can specify. Otherwise I think this is better since we were over constrained below (List[int] when really an iterable or numpy array of integers would be fine).

Mostly just a matter of defining the type, a parsing utility and pushing things through to make mpypy happy
Resolves #305 along the way.

I also realized we didn't set the flag to assert when vals or subs check failed. I turned the flag on but that required fixing some of the tutorial. Since this PR is all about making our interface nicer I think a follow up might be to make vals just be an arbitray OneDArray, I don't think we really need it to be a column array (or if we do we could probably adjust it internally).

----
📚 Documentation preview 📚: https://pyttb--348.org.readthedocs.build/en/348/

### DIFF
--- a/docs/source/tutorial/class_sptensor.ipynb
+++ b/docs/source/tutorial/class_sptensor.ipynb
@@ -271,7 +271,7 @@
    "outputs": [],
    "source": [
     "indices = np.array([[0, 0, 0], [1, 0, 0]])\n",
-    "values = np.ones((2,))\n",
+    "values = np.ones((2, 1))\n",
     "\n",
     "Y = ttb.sptensor.from_aggregator(indices, values)  # Create a sparse tensor.\n",
     "Y"
@@ -300,7 +300,7 @@
    "outputs": [],
    "source": [
     "indices = np.array([[0, 0, 0], [2, 2, 2]])\n",
-    "values = np.array([1, 3])\n",
+    "values = np.array([[1], [3]])\n",
     "\n",
     "Y = ttb.sptensor.from_aggregator(indices, values)  # Create a sparse tensor.\n",
     "Y"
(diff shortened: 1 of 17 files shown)


dmdunla:
@ntjohnson1 It looks like the regression tests passed when this PR was waiting to be merged, but then they failed due to mypy errors when the GitHub actions were run after the PR merge. I confused as to why that would happen. Perhaps I missed a step during the review and merge of the PR. 

I now have broken the dev pipeline pyttb in two different ways today. Ugh!

ntjohnson1:
Between when I posted my PR and you merged it a Numpy release came out which subtly improved their typing support. I fixed it in the other PR so we should be set for the moment.

PR #354, opened 2024-12-07
Explicit layout and fix dense tensor printing

### LINKED ISSUE #338: Display tensors by frontal slice rather than lateral slice
The default display in pyttb is to display the 2D slices w.r.t. the last 2 modes. This is in contrast to MATLAB which displays them w.r.t. the first two modes. For a 3-way tensor, pyttb displays the lateral slices rather than the frontal slices. I think it should be changed or at least there should be a function that displays it the MATLAB TTB way. 

Example in MATLAB:
```
X = tensor([1:8],[2 2 2])
```
Output:
```
X is a tensor of size 2 x 2 x 2
	X(:,:,1) = 
	     1     3
	     2     4
	X(:,:,2) = 
	     5     7
	     6     8
```

Same example in Python:
```
Y = ttb.tensor(np.arange(1,9), (2, 2, 2));
Y
```
Output:
```
tensor of shape (2, 2, 2)
data[0, :, :] =
[[1 5]
 [3 7]]
data[1, :, :] =
[[2 6]
 [4 8]]
```

I expect these display orders are artifacts of the underlying default storage. By default, MATLAB prints in such a way that it's linear in memory. Python uses the opposite ordering by default, so it's perhaps not surprising that its printing is in the opposite order. But since pyttb is storing things in the MATLAB order internally, it should also print them this way.

ntjohnson1:
Fixes #338 and also makes progress towards being more explicit/consistent about our Fortran formatting discussed [here](https://github.com/sandialabs/pyttb/discussions/344) and [here](https://github.com/sandialabs/pyttb/discussions/345#discussioncomment-11261517)

We probably want to capture the fortran ordering in our documentation somewhere and I should probably open an issue to make the ordering flow throughout pyttb so we can eventually support "C" ordering as well.

----
📚 Documentation preview 📚: https://pyttb--354.org.readthedocs.build/en/354/

### DIFF
--- a/pyttb/cp_als.py
+++ b/pyttb/cp_als.py
@@ -185,7 +185,7 @@ def cp_als(  # noqa: PLR0912,PLR0913,PLR0915
 
     # Reduce dimorder to only those modes we will optimize
     dimorder_in = dimorder  # save for output
-    dimorder = [d for d in dimorder if d in optdims]
+    dimorder = [int(d) for d in dimorder if d in optdims]
 
     # Store the last MTTKRP result to accelerate fitness computation
     U_mttkrp = np.zeros((input_tensor.shape[dimorder[-1]], rank))
--- a/pyttb/cp_als.py
+++ b/pyttb/cp_als.py
@@ -92,7 +92,7 @@ def cp_als(  # noqa: PLR0912,PLR0913,PLR0915
      Iter 1: f = ... f-delta = ...
      Final f = ...
     >>> print(M) # doctest: +ELLIPSIS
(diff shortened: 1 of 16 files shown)


dmdunla:
@ntjohnson1 Sorry, I did not fix the merge conflicts correctly after the auto-merge of main into your branch. I tried to make the changes in a Codespace on your fork, but it did not let me push to your branch.

I got confused about the state of the updates, as the conflicts had partial fixes for `order` and `Shape`, and I did not resolve correclty.

Can you please fix this when you get a chence?

PR #359, opened 2024-12-14
Nick/link check

### LINKED ISSUE #358: Check Markdown links GitHub Action has frequent errors that are resolved after 1 or more re-runs
The [GitHub action for checking markdown links](../blob/main/.github/workflows/markdown-check.yml) has non-deterministic behavior. It sometimes fails when trying to reach the [pyttb coveralls site](https://coveralls.io/github/sandialabs/pyttb), but sometimes it doesn't. After re-running the action one or more times, it will eventually pass. 

The link checker has been deprecated and the author now maintains a new link checker. We may want to consider that new one or some other link checker moving forward.

ntjohnson1:
Resolves #358 (if successful)

Also adds [codespell](https://github.com/codespell-project/codespell) to our checks.
Its used in another repo I was contributing to, has 2k stars, runs fast, and is easy to configure. I didn't dig much deeper than that and it caught a bunch of our typos.
I updated the contributing doc, our pre-commit, package tests, and pyproject config to support it. I think that covers all the bases.

----
📚 Documentation preview 📚: https://pyttb--359.org.readthedocs.build/en/359/


ntjohnson1:
The link checker didn't flake on multiple runs. Unfortunately, I wasn't able to figure out how to make the output more verbose, so it's possible I misconfigured it and we are no longer checking links (I don't think that's likely but I can't say 100%)

PR #374, opened 2024-12-28
Draft: Doc update

tgkolda:
Working on various documentation updates. This already has PR #373 embedded, so I'll need to clean it up once that PRs are merged.

### DIFF
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -50,25 +50,15 @@ from pyttb.pyttb_utils import (
 
 
 class tensor:
-    """
-    TENSOR Class for dense tensors.
+    """Class for dense tensors. 
+    
+    **Members**
 
-    Contains the following data members:
+    * ``data``: :class:`numpy.ndarray` containing the data elements of the tensor stored, by default, in Fortran order.
 
-    ``data``: :class:`numpy.ndarray` dense array containing the data elements
-    of the tensor.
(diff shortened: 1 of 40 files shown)


pyttb/tensor.py
```diff
@@ -77,68 +70,118 @@ def __init__(
...
-        shape:
-            Shape of resulting tensor if not the same as data shape.
-        copy:
-            Whether to make a copy of provided data or just reference it.
+        data : optional
+            Source data as :class:`numpy.ndarray` (default: empty).
+        shape : optional
+           Shape of the tensor as a :class:`tuple` (default: ``data.shape()``).
+        copy : optional bool
+            Whether to copy the data or reference it (default: True).
```
ntjohnson1, line 83: Global note: To hopefully avoid wasted effort.

The type hints already capture a most of the additions on these two lines and the type hints can be programmatically verified to make sure they are correct. In the help/generated documentation/ide support these should already be reflected. Both the fact that it is a bool with a default the specified default value.

If that's not coming up on your workflow could you create a discussion with details about how you are interacting with things? I'd like a significant motivation to duplicate information in a way that is more error prone.

dmdunla:
@ tgkolda I'm closing this PR, so if you want to extract parts from it, you'll need to search from the PR page and search or just use the direct URL: https://github.com/sandialabs/pyttb/pull/374

PR #385, opened 2025-01-31
Using matmul instead of @ to enforce order on tenmat product

dmdunla:
----
📚 Documentation preview 📚: https://pyttb--385.org.readthedocs.build/en/385/

### DIFF
--- a/pyttb/tenmat.py
+++ b/pyttb/tenmat.py
@@ -512,7 +512,7 @@ class tenmat:
             if not tshape:
                 return (self.data @ other.data)[0, 0]
             tenmatInstance = tenmat(
-                self.data @ other.data,
+                np.matmul(self.data, other.data, order=self.order),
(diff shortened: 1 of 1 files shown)


ntjohnson1 (approved):
LGTM for this use case. It might make sense for use to globally replace the `@` operator with this more explicit version throughout the project if we hit this issue in the future.

dmdunla:
OK, will merge this now and add Issue to make this consistent across the board later.

ntjohnson1:
Makes sense. I pushed through the bulk of the memory layout stuff in one go and figured there would be some fixes in the tail end required. The loud warnings should help point them out, but won't break functionality. Mostly just highlighting slight inefficiencies.

As we fix the remaining pieces we should get pretty close to supporting C or F order and it'll just be a matter of prioritizing that and making sure our test coverage is appropriate.

Issue #383, opened 2025-01-01
Reshape creates a copy of the data but shouldn't


tgkolda:
The method `tensor.reshape` should not make a deep copy of the data. Either reshape should *change* the tensor it's called on, or it should be a *function* rather than a method. In numpy, the default behavior of reshape is that it returns a view (if possible, depending on the order). I think we get in trouble because we then call the tensor constructor which makes a copy. We need an option to not make a copy. Ideally, we can have two tensors with different shapes that share the same underlying data, but this needs to be clear so that the user knows when they have a view versus a copy.

PR #386, opened 2025-01-31
Change reshape to avoid making a copy of the underlying data

dmdunla:
----
📚 Documentation preview 📚: https://pyttb--386.org.readthedocs.build/en/386/

### DIFF
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -1193,7 +1193,7 @@ class tensor:
             assert False, "Reshaping a tensor cannot change number of elements"
 
         return ttb.tensor(self.data.reshape(shape, order=self.order), shape, copy=False)
-    
+
     def scale(
         self,
         factor: Union[np.ndarray, ttb.tensor],
--- a/pyttb/tensor.py
+++ b/pyttb/tensor.py
@@ -1192,8 +1192,8 @@ class tensor:
         if prod(self.shape) != prod(shape):
             assert False, "Reshaping a tensor cannot change number of elements"
 
-        return ttb.tensor(np.reshape(self.data, shape, order=self.order), shape)
-
+        return ttb.tensor(self.data.reshape(shape, order=self.order), shape, copy=False)
+    
     def scale(
         self,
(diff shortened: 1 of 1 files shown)


dmdunla:
Illustrating that the copy is now avoided ...

Create a `ttb.tensor` and show info regarding the underlying `data` member:
```python
>>> import numpy as np
>>> import pyttb as ttb
>>> A = ttb.tensor(np.array([1,2,3,4]))
>>> np.info(A.data)
class:  ndarray
shape:  (4,)
strides:  (8,)
itemsize:  8
aligned:  True
contiguous:  True
fortran:  True
data pointer: 0x7fbd21913ca0
byteorder:  little
byteswap:  False
type: int64
```

Now reshape it and see that the underlying data pointer is the same after the reshape, with the shape, strides, and contiguous fields changed:

```python
>>> B = A.reshape((2,2))
>>> np.info(B.data)
class:  ndarray
shape:  (2, 2)
strides:  (8, 16)
itemsize:  8
aligned:  True
contiguous:  False
fortran:  True
data pointer: 0x7fbd21913ca0
byteorder:  little
byteswap:  False
type: int64
```

Also, to demonstrate we are referencing the same data across `A` and `B`, we see a change in one changes the other:

```python
>>> A[0] = 2
>>> B
tensor of shape (2, 2) with order F
data[:, :] =
[[2 3]
 [2 4]]
```

dmdunla:
@tgkolda Let me know if this resolves what you were requesting in #383

tgkolda:
@dmdunla - I had envisioned that copy would be a flag (defaulting to false) in the same way it is for numpy's reshape, but I guess that's not critical. Also, the comments should make it clear that it's a reference and not a deep copy.

dmdunla:
Before making the change to include `copy` as an argument to `tensor.reshape`, we should make a plan for how to handle `copy` in all of the data member `__init__` methods so that the use of `copy` is consistent across `pyttb`.

So, I will merge this update for now, which avoids the copy in `reshape` and start a Discussion on `copy`.

PR #389, opened 2025-02-03
Using standard python3 pip installer command

dmdunla:
----
📚 Documentation preview 📚: https://pyttb--389.org.readthedocs.build/en/389/


dmdunla:
@tgkolda Can you verify this change works on Windows?

tgkolda:
Hi Danny, Alas, `python3` is not recognized. Only `python`.

PR #391, opened 2025-02-03
Add 3.13 to regression tests

dmdunla:


dmdunla:
`coveralls` does not yet support Python 3.13, so this is not working. We may need to split the coverage testing away from the full regression testing.

Issue #363, opened 2024-12-17
Example tensor for cp_als needs to have CP structure


tgkolda:
The example being used to demo cp_als in https://pyttb.readthedocs.io/en/latest/tutorial/algorithm_cp_als.html is just a random tensor. Random tensors do not have CP structure, so this is not a good example to use. Either use the amino acids data or create a small tensor with structure.

dmdunla:
@tgkolda To date, we have tried to provide synthetic data in `pyttb` to avoid requiring users to download external data.  So below is an attempt at generating the data, followed by questions on suggestions for the remainder of the tutorial.

Here is a change that creates an example of a small tensor with CP structure:

```python
# Pick the rank and shape
R = 3
tensor_shape = (6, 8, 10)

# Set seed for reproducibility
np.random.seed(0)

# Create a low-rank dense tensor from a Kruskal tensor
factor_matrices = [np.random.rand(s,R) for s in tensor_shape]
X = ttb.ktensor(factor_matrices).to_tensor()
```

Using `cp_als` to compute a rank-R CP decompositions (on the next line in the tutorial) ...

```python
# Compute a solution with final ktensor stored in M1
np.random.seed(0)  # Set seed for reproducibility
short_tutorial = 10  # Cut off solve early for demo
(M1, M1_init, M1_out) = ttb.cp_als(X, R, maxiters=short_tutorial)
```

produces the following output, illustrating that we can recover the structure:
```
CP_ALS:
 Iter 0: f = 1.000000e+00 f-delta = 1.0e+00
 Iter 1: f = 1.000000e+00 f-delta = 1.7e-08
 Final f = 1.000000e+00
```

I like this example because I think that illustrating that `cp_als` can recover the low-rank CP structure almost exactly when knowing the true rank of the input tensor is useful for beginner users. However, the remainder of the tutorial would need to be refactored, as the aim is to illustrate the impact of using different initializations and more computation to achieve a better decomposition.

Questions: 
1. Would adding noise to `X` (maybe as `X_noisy`) suffice to illustrate the concepts in the current tutorial? 
2. Or should we consider some other input tensors to use to illustrate these ideas (e.g., rank-(R+1), then compute rank-R CP decompositions)? 
3. Or should we refactor the whole tutorial to tell a different story?
4. Or should we consider using external data (which currently is available in MATLAB *.mat format, for which loading has not been fully address yet in `pyttb`)?

dmdunla:
[...]

PR #392, opened 2025-02-05
363 example tensor for cp als needs to have cp structure

dmdunla:
@tgkolda Can you review this PR, which addresses #363 ?

----
📚 Documentation preview 📚: https://pyttb--392.org.readthedocs.build/en/392/


dmdunla:
Note that this only changes the tutorial for `cp_als`. Once this looks good, we should make changes for other algorithms to use low-rank tensors as input. Currently, the tutorials were aimed at replication of the MATLAB TTB tutorials, except that loaded data was replaced with generated data. So, this should be revisited to improve consistency across the tutorials.

tgkolda:
Hi @dmdunla, This is a bad example because it takes too long to converge. Using normal random values rather than uniform for factor matrices accelerates the convergence and should obviate the need to the limit of iterations.

PR #416, opened 2025-03-23
Fix #415 - Changing autoclass content to class

### LINKED ISSUE #415: Class description versus constructor (documentation)
The `autoclass_content` setting is currently set to `"both"` in conf.py:
https://github.com/sandialabs/pyttb/blob/f491e1b6638a033901846c721c3d17c49bd8ba95/docs/source/conf.py#L200

The default `autoclass_content` is normally `"class"` (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autoclass_content), and this makes more sense to me. The `"both"` combines the class description with the documentation of `__init__`. 

Setting `autoclass_content` instead to `"class"` (the normal default) splits the documentation of `__init__` off. Especially because our classes have multiple options for construction, I think this makes more sense.

We do need to go back and remove `__init__` from the `:exclude_members:` options in each class rst file.

tgkolda:
Fixes #415. Enables splitting of class description from its `__init__` constructor.

----
📚 Documentation preview 📚: https://pyttb--416.org.readthedocs.build/en/416/


tgkolda:
@dmdunla - The current set of regression test failures seem to be a system error rather than a problem on my side.

tgkolda:
@dmdunla - I don't think this failure is real. Can you resubmit? Thanks.

tgkolda:
@dmdunla - This is the PR about `__init__`

dmdunla:
I really do not want to include the class methods `__init__` in the user documentation.

I think it is confusing to state to users that there is a function called `__init__` and show an example that makes use of it implicilty. For example, in the case of tensor, the documentation proceeds as follows:

1. Provide class name
```
class pyttb.tensor(data: ndarray | None = None, shape: int | Iterable[int]] | None = None, copy: bool = True)
```
2. List `__init__` method
```
__init__(data: ndarray | None = None, shape: int | Iterable[int]] | None = None, copy: bool = True)
```
3. Provide example of using the `__init__` method
```
>>> T = ttb.tensor()
...
```

In #3 above, when instantiating a `tensor` object, `tensor.__new__()` is called to create the object and `tensor.__init__()` is called to initialize the object.

I believe that exposing this subtlety in end-user documentation can be overwhelming and confuse new users who are not familiar with Python classes. Many other packages (used by pyttb or otherwise) add class documentation, with the class constructor API defined when the class is introduced, and do not explicitly provide documentation for `__init__` (or `__new__`, if provided). 
```

tgkolda:
I'm confused a bit by your comment. This is what the documentation looks like with the proposed change. The class does not provide all the input arguments. As for the user confusion, I think this is just a matter of adding explanatory text to `__init__` which is on my to-do list once we get things sorted at this higher level.

image

Issue #418, opened 2025-03-24
Making documentation consistent with namespace promotion.


tgkolda:
Currently, much of the documentation is incorrect. 

The documentation refers to `pyttb.tensor.tensor` as a class, but that won't work!

Instead, it must be `pyttb.tensor` because it has been promoted.

Check the documentation namespace is consistent with the following promotions from:
https://github.com/sandialabs/pyttb/blob/6279f8815029b89a1b21d553396c121b700aff0f/pyttb/__init__.py#L14-L32

- [x] from pyttb.cp_als import cp_als
- [x] from pyttb.cp_apr import cp_apr
- [ ] from pyttb.export_data import export_data
- [x] from pyttb.gcp_opt import gcp_opt
- [x] from pyttb.hosvd import hosvd
- [ ] from pyttb.import_data import import_data
- [ ] from pyttb.khatrirao import khatrirao
- [x] from pyttb.ktensor import ktensor
- [ ] from pyttb.matlab import matlab_support
- [x] from pyttb.sptenmat import sptenmat
- [x] from pyttb.sptensor import sptendiag, sptenrand, sptensor
- [ ] from pyttb.sptensor3 import sptensor3
- [x] from pyttb.sumtensor import sumtensor
- [ ] from pyttb.symktensor import symktensor
- [ ] from pyttb.symtensor import symtensor
- [x] from pyttb.tenmat import tenmat
- [x] from pyttb.tensor import tendiag, teneye, tenones, tenrand, tensor, tenzeros
- [x] from pyttb.ttensor import ttensor
- [x] from pyttb.tucker_als import tucker_als

tgkolda:
The ones that aren't checked are not yet included in sphinx.

dmdunla:
Fixed with 3ba562702277aac44a8630ecd3876206f366adae

Issue #417, opened 2025-03-24
Issues with lifting of classes and functions into pyttb namespace, documentation, etc.


tgkolda:
I have been very confused about the qualification of names of classes and functions within PyTTB until I found these (sneaky) lines of code in the init file:

https://github.com/sandialabs/pyttb/blob/6279f8815029b89a1b21d553396c121b700aff0f/pyttb/__init__.py#L14-L32

The reason I was confused is that the sphinx documentation wants to include the module names in the documentation. However, using the module names will cause PyTTB to break. Observe:

![Image](https://github.com/user-attachments/assets/b486dffc-ad3d-47ea-b3d0-666aeae02573)

This is really confusing!!

I could envision a few possible fixes. I'm guessing the error that I'm having is that by importing tensor directly as a class it no longer makes sense to reference it as a module, so one recommendation is that the modules be named differently than the classes. If they are to be hidden, then maybe an underscore or something.

Then, in the documentation, we need to be ultra careful about the way that the methods are documented. Calling automodule automatically inserts the module name. So, we need to instead use autoclass and autofunction directly.

tgkolda:
As another example, it seems that the function `ttb.cp_apr.vectorize_for_mu` is not accessible because `cp_apr` has been lifted.

tgkolda:
One proposal to consider is naming "hidden modules" as followed by _hm or something. For example, tensor.py would become tensor_hm.py.

ntjohnson1:
One thing I'm partial to is to potentially introduce more nesting. I think we originally flattened things to better match MATLAB.

When I started to play with rust I had the layout look like
```
/ttb
   /tensors
      dense
      sparse
      kruskal
  /cp
     als
```
We can keep the names as is but a different logical grouping would potentially remove some of the confusion about the import paths vs the code paths. Also importing everything in the flat layout has made a few things minorly inconvenient because we had to be more careful about import ordering.

dmdunla:
Closing, as this is superceded by Issue #422 / PR #444

PR #421, opened 2025-03-24
Fixing #418 - Resolving namespace issues in documentation.

tgkolda:
Fixing automodule versus autoclass or autofunction for classes and functions that are promoted to the pyttb level. Also adding notes to explain the promotion which is otherwise hidden from the user.

This mostly addresses #417 as well. I should have made #418 a subissue!

----
📚 Documentation preview 📚: https://pyttb--421.org.readthedocs.build/en/421/

### DIFF
--- a/docs/source/cpals.rst
+++ b/docs/source/cpals.rst
@@ -1,7 +1,8 @@
 CP Alternating Least Squares (:obj:`cp_als`)
 ============================================
+.. note::
+    
+     The ``cp_als`` function defined in ``cp_als.py`` has been promoted to the ``pyttb`` namespace.
 
-.. automodule:: pyttb.cp_als
-    :members:
-    :undoc-members:
-    :show-inheritance:
+.. autofunction:: pyttb.cp_als
+    
--- a/docs/source/cpapr.rst
+++ b/docs/source/cpapr.rst
@@ -1,7 +1,8 @@
 CP Alternating Poisson Regression (:obj:`cp_apr`)
 =================================================
+.. note::
+    
+     The ``cp_apr`` function defined in ``cp_apr.py`` has been promoted to the ``pyttb`` namespace.
 
-.. automodule:: pyttb.cp_apr
-    :members:
-    :undoc-members:
-    :show-inheritance:
+.. autofunction:: pyttb.cp_apr
+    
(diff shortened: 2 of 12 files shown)


dmdunla:
I really do not want to include the class methods `__init__` in the user documentation.

I think it is confusing to state to users that there is a function called `__init__` and show an example that makes use of it implicilty. For example, in the case of tensor, the documentation proceeds as follows:

1. Provide class name
```
class pyttb.tensor(data: ndarray | None = None, shape: int | Iterable[int]] | None = None, copy: bool = True)
```
2. List `__init__` method
```
__init__(data: ndarray | None = None, shape: int | Iterable[int]] | None = None, copy: bool = True)
```
3. Provide example of using the `__init__` method
```
>>> T = ttb.tensor()
...
```

In #3 above, when instantiating a `tensor` object, `tensor.__new__()` is called to create the object and `tensor.__init__()` is called to initialize the object.

I believe that exposing this subtlety in end-user documentation can be overwhelming and confuse new users who are not familiar with Python classes. Many other packages (used by pyttb or otherwise) add class documentation, with the class constructor API defined when the class is introduced, and do not explicitly provide documentation for `__init__` (or `__new__`, if provided). 
```

dmdunla:
Evolving the documentation incrementally in the way we have been doing in the past few merges seems to be causing challenges to progressing on overall development (e.g., I'd like to release a new patch version to support several users but I do not want to release the documentation in a mid-revision state). This is my fault, and I take full responsibility. I had thought the doc change requests were relatively minor, but as the discussion has progressed, I see that it is more complicated in several ways.

I'd like to move documentation aside from the main development, creating a new branch dedicated to a comprehensive documentation overhaul that would then be merge altogether into a new minor release (rather than as patches as we have been doing).

Here are some things that we should discuss and work on to come to a useful set of docs:
1. Who is the audience of the RTD docs?
  - My suggestion is the end user. We could provide more docs for developers, but I believe this is a lower priority right now. And I do not want to conflate the two in a single set of docs.
2. How do we want to document a class?
  - My suggestion is to follow conventions used by other (major) Python packages that users of `pyttb` might use -- e.g., `numpy`, `scipy`, `pandas`.
  - This includes some standardization of the class doc format, too.
  - This includes determining which methods to include -- e.g., dunder (`__foo__`), internal (`_bar`), and helper functions included in module files but not specified as methods of the main class defined in that module file.
3. What are the entry points we expect for documentation?
  - readthedocs.io
  - `help(pyttb)`
  - Reading through the source code
4. What should the interplay between the docs and the tutorial be?
  - Are there different audiences for these?
  - How should we align content/examples/tests?

PR #423, opened 2025-04-04
Updates to tensor class documentation (only at module and class level)

tgkolda:
- Move assumed module imports into `tensor.rst` so that it appears at the top of the document.
- Remove "TENSOR" in the class description, which is a remnant from MATLAB.
- Document the attributes of the class, which before were only partly documented.
- Document all the ways that a tensor can be created.

_Note:_ This does not depend on the other PRs. There may be some minor issue with `tensor.rst` depending on the order of the merges. If there is, let me know and I can fix.  There may be possible issues with crosslinks until #421 is merged, but this is sort of a pre-existing problem.

----
📚 Documentation preview 📚: https://pyttb--423.org.readthedocs.build/en/423/

### DIFF
--- a/docs/source/tensor.rst
+++ b/docs/source/tensor.rst
@@ -1,6 +1,11 @@
 Dense Tensor (:class:`tensor`)
 ------------------------------
 
+For *all* examples in this document, the following module imports are assumed::
+
+    >>> import pyttb as ttb
+    >>> import numpy as np
+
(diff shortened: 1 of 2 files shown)


tgkolda:
Direct link to modified documentation: https://pyttb--423.org.readthedocs.build/en/423/tensor.html

Issue #445, opened 2025-08-28
Remove Python 3.9 and add Python 3.14


ntjohnson1:
Deprecation/ Release in October. When we do this I think we should add the TC and UP rulesets from ruff to make sure we are up to date with our minimum python style. They should generate some churn but be mostly automatic.

PR #446, opened 2025-08-29
Update Ruff

ntjohnson1:
This is a lot of churn but basically all automatic with ruff check --fix. Since we've cleared the backlog probably nice to do it before there are too many things in flight.

Changes:
* Ruff 0.7 -> 0.12
* Added rules:
   * https://docs.astral.sh/ruff/rules/#pyupgrade-up
       * Mostly just make sure we are aligned with the latest coding standards for our minimum python
   * https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc
      * Helps a bit with some circular imports and reducing overhead for type checking stuff not needed at runtime
   * https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa
      * Adding the future annotations in 3.9/3.10 enables more features so turn it on by default and enforce it with this

----
📚 Documentation preview 📚: https://pyttb--446.org.readthedocs.build/en/446/


dmdunla:
@ntjohnson1 In #445, you mention adding the updated ruff rules at the release/deprecation changeover point. Just making sure, you'd like this to land now rather than waiting for that (in October).

dmdunla:
Although this could land now, I am going to wait until after the next release, since it touches almost everything. The biggest thing is that it significantly changes out use of typing across the board, and I think it would be better to make this change when we finalize the doc structure in the following release.

dmdunla:
> > make this change when we finalize the doc structure in the following release.
> 
> Sounds good. I'll close this then

Sorry, I meant I DO want this to land this next, before we land the doc updates in the next release. The changes to the typing alone really helps improve readability of the docs.

So, I'm re-opening this PR and plan to merge next.

PR #450, opened 2025-08-31
311 implement zero truncated poisson algorithm

### LINKED ISSUE #311: Implement zero-truncated Poisson algorithm
This issue will start out as a catch-all for implementing ZTP (see [listings 1--3](https://doi.org/10.1093/imaiai/iaad016)). We anticipate several future issues (e.g., visualization, ensuring the interface of some untested `gcp_opt` functionality) that will be linked to the eventual PR.

jeremy-myers:
Closes #311 

The tutorial is a minimal port of Listing 2 from (https://academic.oup.com/imaiai/article/12/3/1573/7159150#417226967).

----
📚 Documentation preview 📚: https://pyttb--450.org.readthedocs.build/en/450/

### DIFF
--- a/tests/gcp/test_optimizers.py
+++ b/tests/gcp/test_optimizers.py
@@ -35,7 +35,7 @@ def generate_problem():
 def test_sgd(generate_problem):
     dense_data, model, sampler = generate_problem
 
-    solver = SGD(max_iters=2, epoch_iters=1)
+    solver = SGD(max_iters=3, epoch_iters=1)
     result, info = solver.solve(
         model, dense_data, gaussian, gaussian_grad, sampler=sampler
     )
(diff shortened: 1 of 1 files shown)


dmdunla:
@jeremy-myers Looks like this is failing regressions due to some slight numerical differences in the expected test output. Can you take a look and see if Nick's example of using `np.allclose()` shown [here](https://github.com/sandialabs/pyttb/blob/c6d5fe2639d66ffa85ff8c38aac569220d287534/tests/test_ktensor.py#L681) could work for the ZTP tests you provided?

jeremy-myers:
> @jeremy-myers Looks like this is failing regressions due to some slight numerical differences in the expected test output. Can you take a look and see if Nick's example of using `np.allclose()` shown [here](https://github.com/sandialabs/pyttb/blob/c6d5fe2639d66ffa85ff8c38aac569220d287534/tests/test_ktensor.py#L681) could work for the ZTP tests you provided?

@dmdunla This is failing on `tests/gcp/test_optimizers.py::test_sgd` which doesn't make sense to me - this PR doesn't touch that test. I don't have a debugger set up at the moment - can you take a look?

PR #467, opened 2025-11-14
Binary export/import

### LINKED ISSUE #466: Should we support a binary serialization format?
Follow on from: https://github.com/sandialabs/pyttb/issues/460

My current opinion is that since python has pickle we don't need our own backwards compatible binary format.

Pro/Con for binary format

Pros:
* Isn't tied to python version/pyttb version
* Is probably dramatically faster than our current text based format

Cons:
* Maintenance
* Would probably be really hard to make compatible with MATLAB ttb

ntjohnson1:
This isn't fully ready to merge since we need cross repo collaboration but I think it is close enough where concrete feedback is useful. There are also a few minor TODOS (adding a version to our headers etc). There are a few open questions around how to test this, and if people like the overall api.

Here is my branch for the matching half in MATLAB: https://gitlab.com/ntjohnson1/tensor_toolbox/-/commits/nick/import_export_compatibility (I figure we should make sure we like it here before opening discussion there)

General usage looks like:
MATLAB export
```matlab
ktensor_instance = <value>;
export_data('ktensor.mat', 'output_type', 'binary');
```
Python import
```python
import pyttb as ttb
ktensor_instance = ttb.import_data_mat('ktensor.mat')
```
OR
Python export
```python
import pyttb as ttb
ktensor_instance = <value>
ttb.export_data_mat(ktensor, 'ktensor.mat')
```
MATLAB import
```matlab
ktensor_instance = import_data('ktensor.mat', 'input_type', 'binary');
```

NOTE: I left `import_data_bin` and `export_data_bin` on our side since its not much overhead and if you aren't using matlab it seems a little strange to write to `.mat`.

Closes #466 

Testing options:
* Check-in an export file from the respective other library in each repo and make a note to re-export and upload if changing format
   * On the pyttb side we can add a lint in CI to enforce this
* Setup matlab in CI (on either repo) to actually run the compatibility test
   * If MATALB tensortoolbox already had this running python is so much easier, but seems like a lot of effort to setup just for this 
* Other ideas?

Remaining steps:
* [ ] Get initial review on the Python side
* [ ] Get initial review on the MATLAB side
* [ ] Merge one then the other

----
📚 Documentation preview 📚: https://pyttb--467.org.readthedocs.build/en/467/

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,9 @@
 # U.S. Government retains certain rights in this software.
 from __future__ import annotations
 
+import os
+import tempfile
+
 import numpy
 import numpy as np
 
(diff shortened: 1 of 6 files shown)


gqcollins:
I'll try it next week and get back to you. Thanks!

